Issue #7, opened 2026-05-14
Lib unit tests broken on main: stale Environment trait usage + missing types


rjwalters:
## Summary

\`cargo test --lib --features training --no-run\` fails with **42 compile errors** on a fresh checkout of \`main\` (commit \`f9b3779\`). The library itself compiles cleanly (\`cargo check\` passes); the failure is confined to in-module \`#[cfg(test)]\` blocks.

## Repro

\`\`\`
$ cd thrust
$ export LIBTORCH=$PWD/libtorch  # or your install
$ cargo test --lib --features training --no-run 2>&1 | grep -c "^error"
42
\`\`\`

## Affected files

The first ~10 errors localize to two files:

\`\`\`
src/multi_agent/simulator.rs:270-305  -- uses Environment::Observation, Environment::Action
                                         (no such associated types on the current trait)
                                      -- impl Environment for ... missing get_observation, render, close
src/inference/weights.rs:60-61        -- references undeclared type \`LayerWeights\`
\`\`\`

## Root cause

Looks like the \`Environment\` trait was refactored from an associated-type style to the flat \`StepResult\`-returning style that's currently in \`src/env/mod.rs\`, but the in-module test code in \`multi_agent/simulator.rs\` was missed. Same for \`LayerWeights\` --- presumably renamed or removed.

## Impact

- Any new contributor who runs \`cargo test\` is greeted with 42 errors that aren't their fault.
- Adding a new in-module unit test requires either fixing all 42 errors first (out of scope for a single PR), or putting the test in \`tests/\` as a separate integration binary.

## Suggested fix

Either:
1. Bring the test code in \`multi_agent/simulator.rs\` up to the current trait shape.
2. Or temporarily \`#[ignore]\`/\`#[cfg(skip_broken)]\` the offending tests with a TODO referencing this issue, so the rest of the suite is runnable.

Happy to do #1 in a follow-up PR if there's appetite.

rjwalters:
## Curator enrichment (2026-05-14)

Verified on `master` @ `7ba44e3` with `LIBTORCH=$PWD/libtorch cargo test --lib --features training --no-run`. **42 compile errors confirmed.** `cargo check --lib --features training` passes cleanly — failures are entirely confined to `#[cfg(test)] mod tests { ... }` blocks.

### Confirmed scope: this is broader than the original report

The issue body lists `src/multi_agent/simulator.rs` and `src/inference/weights.rs` as the two affected files. The actual breakdown is wider:

[...]

Issue #2, opened 2026-05-14
PPOTrainer::train_step: support an auxiliary loss term


rjwalters:
## Motivation

\`PPOTrainer::train_step\` currently composes its total loss as

\`\`\`
total = policy_loss + value_coef * value_loss - entropy_coef * entropy
\`\`\`

with no way for the caller to inject an additional term. Many algorithms need this:

- **Cross-agent redundancy penalty** (Slepian-Wolf MARL, P3 specialization): \`+ λ * Σ_{i<j} ||corr(Ẑ_i, Ẑ_j)||_F^2\`
- **Behavioural diversity** (DIAYN-style): \`- α * H(skill | trajectory)\`
- **Imitation regularizer**: \`+ β * KL(π_θ || π_demo)\`
- **Representation auxiliary**: \`+ γ * MSE(predicted_next_state, actual_next_state)\`

## Current API

\`\`\`rust
pub fn train_step<F>(
    &mut self,
    observations: &Tensor, actions: &Tensor,
    old_log_probs: &Tensor, old_values: &Tensor,
    advantages: &Tensor, returns: &Tensor,
    mut forward_fn: F,
) -> Result<TrainingStats>
where
    F: FnMut(&Tensor, &Tensor) -> (Tensor, Tensor, Tensor),
\`\`\`

## Proposal

Add an optional aux-loss callback:

\`\`\`rust
pub fn train_step_with_aux<F, A>(
    &mut self,
    // ...same args...
    mut forward_fn: F,
    mut aux_fn: A,
) -> Result<TrainingStats>
where
    F: FnMut(&Tensor, &Tensor) -> (Tensor, Tensor, Tensor),
    A: FnMut(&Tensor /* mb_obs */) -> Option<Tensor>, // scalar loss to add
\`\`\`

Or pass an extra \`aux_loss: Option<Tensor>\` along with \`(log_probs, entropy, values)\` from \`forward_fn\` (cleaner — keeps the loss computation co-located with the policy forward).

Either way, \`TrainingStats\` should grow an \`aux_loss\` field for logging.

## Notes

This is the natural counterpart to #1 (encoder tap): the tap exposes the features, this hook exposes the loss slot.

Backwards-compat path: keep \`train_step\` as-is, add \`train_step_with_aux\` and have the original delegate.

rjwalters:
## Curator Note: Appears Already Implemented

[...]

PR #12, opened 2026-05-14
Fix multi_agent::PolicyLearner TODO/placeholder defects

### LINKED ISSUE #4: multi_agent::PolicyLearner has unresolved TODO placeholders
## Summary

## Specifics

\`\`\`rust
    config.buffer_size,
    1, // num_envs
);
\`\`\`

\`\`\`rust
// 3. Compute advantages
// self.config.gae_lambda);
    &last_values,
    self.config.gamma as f32,
);
\`\`\`

## Recommended fix

## Repro

rjwalters:
## Summary

Closes #4

Replace the four TODO/placeholder defects in `src/multi_agent/learner.rs` so the learner produces a correct PPO training signal instead of silently spinning as a no-op. The curator's review of issue #4 identified four defects (not the two originally reported); all four are addressed here.

### Defects fixed

1. **`obs_dim` hardcoded to 4** (`learner.rs:73`) — wrong for every environment except CartPole. Added `obs_dim` to `LearnerConfig` (default `4` for CartPole backward compatibility) and threaded it through to the `RolloutBuffer` constructor.
2. **GAE `last_values` was always `vec![0.0]`** (`learner.rs:108`) — systematically biased advantages downward on every non-terminal rollout end. New `compute_bootstrap_last_values` helper runs the current policy on the final `next_observation` under `no_grad` and returns `V(s_{T+1})`, falling back to `0.0` only when the rollout ends on a terminal transition.
3. **Stale `buffer.capacity()` TODO** (`learner.rs:148`) — removed. Rollout fill level is now tracked locally via `buffer_step_idx` because `RolloutBuffer::len()` returns capacity, not fill.
4. **Most severe: `collect_experiences` was a no-op** (`learner.rs:154-160`) — incoming `Experience` values were bound as `_exp` and discarded; `self.buffer.add(...)` was commented out. The `min_batch_size` guard therefore never tripped and `train()` never ran. New `ingest_experience` helper translates each `Experience` into a `RolloutBuffer::add(...)` call (converting tensor observations to `Vec<f32>` after moving to CPU) and the batch-readiness guard now uses `buffer_step_idx`.

### Regression test

Added `test_learner_ingests_experiences_and_trains` which:
- Constructs a learner with a non-trivial `obs_dim = 8` (not 4 — exercises defect #1).
- Pushes 6 `Experience` values through the channel with the last one terminal.
- Verifies `buffer_step_idx == n_exp` after `collect_experiences` (defect #4: buffer was previously never populated).
- Verifies `compute_bootstrap_last_values()` returns `vec![0.0]` for a terminal end and a finite policy-derived value for a non-terminal end (defect #2).
- Runs one full `compute_advantages` + `train_step` cycle end-to-end and asserts the loss components are finite.

### Other changes

- README: noted `multi_agent::PolicyLearner` is experimental under the Phase 2 status section.

## Test plan

- [x] `cargo build` (passes — lib compiles cleanly with all warnings unrelated to this change).
- [ ] `cargo test --lib multi_agent::learner` — blocked by pre-existing lib-test compile failures unrelated to this change (issue #7); the learner module itself produces no compile errors.
- [ ] Manual smoke test on a multi-agent training run with an env where `obs_dim != 4` (e.g. Snake).

### Notes for the judge

- The library build is clean. The pre-existing failure to compile `--lib` tests (issue #7, missing `pong.rs` referenced by `src/env/games/mod.rs` and various `Environment` trait drift in `src/env/games/cartpole.rs`, `src/buffer/rollout/storage.rs`, `src/multi_agent/simulator.rs`, `src/multi_agent/environment.rs`) prevents running the regression test in CI right now, but is out of scope for this issue.
- No changes were made outside `src/multi_agent/learner.rs` and `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ See [ROADMAP.md](ROADMAP.md) for detailed development schedule.
 - [x] SimpleBandit environment (contextual bandits)
 
 ### Phase 2: Multi-Agent & WASM (In Progress - 60%)
-- [x] Multi-agent training infrastructure
+- [x] Multi-agent training infrastructure (`multi_agent::PolicyLearner` is experimental — API may change)
 - [x] Population-based training design
 - [x] Pure Rust inference (no PyTorch in production)
 - [x] Universal inference system (JSON model format)
--- a/src/multi_agent/learner.rs
+++ b/src/multi_agent/learner.rs
@@ -6,7 +6,7 @@ use std::time::Duration;
 
 use anyhow::Result;
 use crossbeam_channel::{Receiver, Sender};
-use tch::Tensor;
+use tch::{Tensor, no_grad};
 
 use super::{
     messages::{Experience, PolicyUpdate, TrainingStats},
@@ -45,6 +45,30 @@ pub struct PolicyLearner {
 
     /// Path for saving models
     model_save_dir: String,
+
+    /// Index into the rollout buffer for the next experience to be inserted.
+    /// Wraps modulo `config.buffer_size` and is reset to 0 after each
+    /// `train()` cycle (when the buffer is cleared via `reset`).
+    buffer_step_idx: usize,
+
+    /// Snapshot of the most recently received experience for the current
+    /// rollout. Used to compute the GAE bootstrap value `V(s_{T+1})` when
+    /// the rollout ends in a non-terminal state (e.g. truncation or buffer
+    /// full). `None` when the buffer is empty.
+    last_experience: Option<LastExperience>,
+}
+
+/// Minimal information about the most recent experience needed to compute
+/// the GAE bootstrap value on a non-terminal rollout boundary.
+struct LastExperience {
+    /// Flattened next observation for the final inserted transition.
+    next_observation: Vec<f32>,
+
+    /// Whether the final inserted transition terminated the episode.
+    /// A terminal end uses `0.0` as the bootstrap value; a non-terminal
+    /// end (truncation / buffer-full) requires `V(s_{T+1})` from the
+    /// current policy.
+    terminated: bool,
 }
 
 impl PolicyLearner {
@@ -70,7 +94,7 @@ impl PolicyLearner {
         let buffer = RolloutBuffer::new(
             config.buffer_size,
             1, // num_envs
-            4, // obs_dim - TODO: make configurable
+            config.obs_dim,
         );
 
         Ok(Self {
@@ -82,6 +106,8 @@ impl PolicyLearner {
             config,
             step: 0,
             model_save_dir,
+            buffer_step_idx: 0,
+            last_experience: None,
         })
     }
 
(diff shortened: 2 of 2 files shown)


rjwalters:
## Judge Review — Approved

Reviewed all four defects against the curator's acceptance criteria. All addressed correctly.

### Defect verification

1. **`obs_dim` hardcoded (defect #1)** — Fixed at `learner.rs:97`. `LearnerConfig.obs_dim` field added with a CartPole-compatible default of 4, threaded into `RolloutBuffer::new`. The regression test uses `obs_dim = 8` and would trigger the `debug_assert_eq!(observation.len(), self.obs_dim)` in `RolloutBuffer::add` (`src/buffer/rollout/storage.rs:121`) if the hardcoded `4` remained, so this is a genuine regression guard.

2. **GAE bootstrap stuck at `vec![0.0]` (defect #2)** — Fixed via new `compute_bootstrap_last_values()` (lines 272-289). Correctly returns `vec![0.0]` for terminal ends and `None` defaults; otherwise runs `policy.forward(&next_obs)` under `tch::no_grad` and extracts `V(s_{T+1})` via `double_value(&[])`. The `[1]`-shape value-tensor unwrap pattern matches the existing `simulator.rs:162` precedent and is supported by tch's "tensors holding a single element" contract.

3. **Stale `buffer.capacity()` TODO (defect #3)** — Fixed by introducing local `buffer_step_idx`. The diagnosis is correct: `RolloutBuffer::len()` (`storage.rs:149-151`) returns `num_steps * num_envs`, i.e. capacity, so using it as a fill-level counter would never have tripped the `min_batch_size` guard.

4. **Silent no-op `_exp` drop (defect #4 — the critical one)** — Fixed via new `ingest_experience()` (lines 220-262). Converts the observation/next-observation tensors to `Vec<f32>` (with a defensive CPU `to_device` to handle GPU tensors), validates obs dim, calls `RolloutBuffer::add(...)` with the correct 9-arg signature, updates the `LastExperience` snapshot, and advances `buffer_step_idx`. Error handling is propagated as `anyhow::Result`.

### Regression test (`test_learner_ingests_experiences_and_trains`)

Genuinely exercises all four defects end-to-end:
- Uses `obs_dim = 8` (not 4) — defect #1.
- Pushes 6 `Experience` values, drops the sender so `collect_experiences` returns cleanly on timeout, then asserts `buffer_step_idx == 6` and `last_experience.is_some()` — defect #4.
- Asserts `compute_bootstrap_last_values()` returns `vec![0.0]` for the terminal final transition, then forces a non-terminal `LastExperience` and asserts the result is finite — defect #2.
- Runs `compute_advantages` + `train_step` and asserts all loss components are finite.

The defect #3 assertion is implicit (would have panicked the buffer's debug_assert if not fixed) — fine, given that #1 already exercises the dim plumbing.

### CI status (UNSTABLE — explained)

All CI failures are confirmed pre-existing infrastructure issues, not regressions in this PR:
- The `bucket-brigade-core` git submodule fails to initialize in CI runners, breaking every workflow including on `main` (run 25884535543 shows the identical failure on the master branch).
- The lib-test compile errors (42 of them) tracked as issue #7 are in `cartpole.rs`, `simulator.rs`, `env/mod.rs`, `snake.rs`, `policy/inference.rs`, `ppo/loss.rs`, etc. — none in `multi_agent/learner.rs`.
- The uncommitted `src/env/games/pong.rs` is referenced by `env/games/mod.rs` but does not exist in any committed tree (master or this PR).

Locally `cargo build` succeeds cleanly with `LIBTORCH_USE_PYTORCH=1 LIBTORCH_BYPASS_VERSION_CHECK=1`. The learner module itself produces zero compile errors. The PR description honestly discloses both caveats.

### Scope

Diff is exactly 2 files: `learner.rs` (+270/-22) and `README.md` (+1/-1, adds "experimental" qualifier). No scope creep.

### Minor follow-ups (non-blocking)

- The docstring on `buffer_step_idx` (line 50) says "Wraps modulo `config.buffer_size`" but the code never wraps — it stops at capacity and resets to 0 after a training cycle. Cosmetic.
- `train_step()` still consumes the full buffer capacity via `buffer.get_batch()`, so when `min_batch_size < buffer_step_idx < buffer_size`, the trailing zero-padded rows contaminate the batch. This is pre-existing and orthogonal to issue #4 — worth a separate follow-up issue.
- The `unsafe` pointer aliasing in `train_step()` (lines 321-337) is also pre-existing and untouched here.

### Verdict

Approve. Correct, well-tested, well-documented, honest about caveats, and tightly scoped to the four curator-identified defects.

PR #13, opened 2026-05-14
fix(build): keep libtorch_cuda in NEEDED, drop LD_PRELOAD workaround (#9)

### LINKED ISSUE #9: torch_cuda link-lib request stripped by linker; LD_PRELOAD required at runtime to enable CUDA
## Summary

Building thrust with `LIBTORCH_USE_PYTORCH=1` against a pip-installed PyTorch causes the resulting binary to silently run on CPU even when `torch.cuda.is_available()` is true and the build script emitted `cargo:rustc-link-lib=torch_cuda`. The CUDA backend never initializes, so `Device::cuda_if_available()` returns `Cpu`.

`LD_PRELOAD`'ing `libtorch_cuda.so` and `libc10_cuda.so` at run time forces initialization and the binary then reports `Cuda(0)`. But this is a foot-gun and the failure mode (silent CPU fallback) is hard to notice in benchmarks.

## Repro

On a Linux box with PyTorch >= 2.9 installed via pip and a working NVIDIA driver:

```bash
export LIBTORCH_USE_PYTORCH=1
export LIBTORCH_BYPASS_VERSION_CHECK=1
cargo build --release --example train_p3 --features "training env-bucket-brigade"
ldd target/release/examples/train_p3 | grep torch_cuda  # nothing
```

The build script's `output` file says `cargo:rustc-link-lib=torch_cuda`, but `ldd` shows only `libtorch_cpu.so` and `libc10.so` in the runtime deps. This is the linker's `--as-needed` behavior stripping libs that have no referenced symbols at link time.

## Workaround

```bash
TORCH_LIB=$(python3 -c "import torch, os; print(os.path.join(os.path.dirname(torch.__file__), 'lib'))")
export LD_PRELOAD="$TORCH_LIB/libtorch_cuda.so:$TORCH_LIB/libc10_cuda.so"
./target/release/examples/train_p3 ...
```

After this, `device=Cuda(0)` and training is in fact on the GPU.

## Suggested fixes

1. Have the torch-sys build script emit `cargo:rustc-link-arg=-Wl,--no-as-needed` (or `-Wl,--no-as-needed -ltorch_cuda -Wl,--as-needed` pattern) when CUDA is available, so the linker keeps the unreferenced `libtorch_cuda` dependency.
2. Or, have `tch` itself reference at least one symbol from `libtorch_cuda` (e.g. `at::cuda::is_available()`) so `--as-needed` doesn't strip it.
3. Or, document the `LD_PRELOAD` workaround prominently in `GPU_SETUP.md` so users don't silently fall back to CPU.

## Impact

Discovered while benchmarking thrust against the Python pipeline. Without this fix, the touted "3-6x speedup over Python" can collapse to roughly 1x (CPU vs CPU) because the Rust binary never actually uses the GPU.

## Related

* #4 (multi_agent PolicyLearner TODOs)
* #8 (libtorch dylib mismatch on macOS) --- same family of "tch is brittle against the libtorch we have"

---

### Root Cause Analysis

This bug is a confluence of three behaviors:

1. **`torch-sys` build script** correctly emits `cargo:rustc-link-lib=torch_cuda` when it detects a CUDA-enabled PyTorch (verified in our setup via `LIBTORCH_USE_PYTORCH=1` + cu128 wheel).
2. **GNU `ld` defaults to `--as-needed`** on modern Linux distros (Ubuntu 22.04+, Debian 12+, etc.). Under this flag, the linker drops `DT_NEEDED` entries for shared libraries from which no symbols are referenced. Because `thrust` only calls `tch` high-level APIs and `tch` itself only references `libtorch_cpu` / `libc10` symbols directly, the `libtorch_cuda` link request is silently dropped during the final link of the binary.
3. **`libtorch` performs lazy CUDA backend registration**: it only initializes the CUDA backend if `libtorch_cuda.so` (and `libc10_cuda.so`) is *loaded into the process address space*. There is no runtime `dlopen` fallback — `libtorch` checks the global symbol table on init. So a binary with no `DT_NEEDED libtorch_cuda.so` entry will return `false` from `tch::Cuda::is_available()` even on a CUDA-equipped box.

The runtime failure mode is *silent*: `Device::cuda_if_available()` returns `Device::Cpu` without warning, training proceeds on CPU at ~1/5th the expected throughput, and only careful benchmarking reveals it.

**Note on the existing `--no-as-needed` workaround**: `scripts/train-gpu.sh:59` and `scripts/gpu-train.sh:38` already set `RUSTFLAGS="-C link-arg=-Wl,--no-as-needed"`. In practice this is **insufficient by itself** — `docs/GPU_SETUP.md:36-40` documents that "the linker drops CUDA library as 'unused' even with `--no-as-needed`" and the `LD_PRELOAD` step is still required. The likely reason is that `RUSTFLAGS` `-C link-arg` adds the flag at a position in the linker command line that does not cover the `-ltorch_cuda` argument emitted by `torch-sys` (linker flag scope is positional: `--no-as-needed` only affects libraries *after* it on the command line, until `--as-needed` is seen). The fix requires the flag to be emitted *by the build script* immediately before `-ltorch_cuda`.

### Implementation Options

**Option A: Upstream fix in `torch-sys` build.rs (preferred, but external dependency)**

Patch `torch-sys` to emit positional `cargo:rustc-link-arg` directives that bracket the CUDA libs:

```rust
// In torch-sys/build.rs, when use_cuda is true:
println!("cargo:rustc-link-arg=-Wl,--no-as-needed");
println!("cargo:rustc-link-lib=torch_cuda");
println!("cargo:rustc-link-lib=c10_cuda");
println!("cargo:rustc-link-arg=-Wl,--as-needed");
```

This guarantees the CUDA libs survive `--as-needed` regardless of downstream RUSTFLAGS positioning. Requires a PR to https://github.com/LaurentMazare/tch-rs and a version bump. **Out of scope for this repo** in the short term.

**Option B: Force-reference a CUDA symbol from thrust's own code (in-repo, recommended)**

Add a `#[used]` static or a deliberate call to `tch::Cuda::is_available()` early in initialization, *inside a function the linker cannot eliminate*. The trick is that `tch::Cuda::is_available()` ultimately resolves to a symbol in `libtorch_cuda` only when CUDA support is compiled in; otherwise it stubs to false. Adding it in main and gating with the `training` feature defeats `--as-needed` for CUDA-enabled builds:

```rust
// src/lib.rs or a new src/cuda_anchor.rs (feature = "training" only)
#[cfg(feature = "training")]
#[allow(dead_code)]
#[doc(hidden)]
pub fn _cuda_link_anchor() -> bool {
    // Touching this symbol drags libtorch_cuda into the link.
    tch::Cuda::is_available()
}
```

Then call `crate::_cuda_link_anchor()` from a `static` initializer or a stub used by every `Device::cuda_if_available()` callsite. **Caveat**: `tch::Cuda::is_available()` may not pull a `libtorch_cuda`-only symbol; needs experimental verification (build, then `ldd target/.../bin | grep cuda` and `readelf -d target/.../bin | grep NEEDED`).

A more reliable variant is to bind a known CUDA-only `extern "C"` symbol via `extern "C" { fn at_cuda_synchronize(); }` (or similar from `c10::cuda` namespace) and reference its address — `#[used] static _ANCHOR: unsafe extern "C" fn() = at_cuda_synchronize;` — gated behind `cfg(feature = "training")` and possibly a new `cfg(thrust_cuda)` flag set by a thin `build.rs` that mirrors `torch-sys`'s CUDA detection.

**Option C: Custom build.rs emitting positional link args (in-repo, robust)**

Add a top-level `build.rs` to `thrust-rl` that:

1. Detects whether CUDA is enabled (mirror `torch-sys` detection logic, or read `cargo:rustc-cfg=use_cuda` cascaded from `torch-sys` if available).
2. Emits, when CUDA is on:
   ```
   cargo:rustc-link-arg=-Wl,--no-as-needed
   cargo:rustc-link-lib=torch_cuda
   cargo:rustc-link-lib=c10_cuda
   cargo:rustc-link-arg=-Wl,--as-needed
   ```

`cargo:rustc-link-arg` from a leaf-crate build.rs is appended *after* dependency link args, so the positional `--no-as-needed` will appear at the end of the link command — covering our explicit re-emission of `-ltorch_cuda`. This is the most surgical fix that doesn't depend on upstream `tch-rs`.

**Option D: Status quo + better detection (fallback)**

Keep the `LD_PRELOAD` workaround in scripts, but add a startup check that hard-errors if CUDA-flagged builds end up on CPU:

```rust
// In examples that expect GPU:
let device = Device::cuda_if_available();
#[cfg(feature = "training")]
if std::env::var("THRUST_EXPECT_CUDA").is_ok() && device == Device::Cpu {
    eprintln!("FATAL: THRUST_EXPECT_CUDA was set but tch fell back to Cpu.");
    eprintln!("Hint: ensure LD_PRELOAD includes libtorch_cuda.so + libc10_cuda.so.");
    std::process::exit(2);
}
```

Plus update `docs/GPU_SETUP.md` and `docs/GPU_TRAINING_GUIDE.md` to mark the `LD_PRELOAD` step as **REQUIRED**, not just helpful.

### Recommendation

Pursue **Option C** as the primary fix (in-repo `build.rs` with positional link args), with **Option D** as a hardening guard. Option C should make `cargo build` produce a binary that links `libtorch_cuda` correctly without any runtime `LD_PRELOAD` gymnastics. Option D ensures we fail loudly if Option C ever regresses or if the user is on a setup where it doesn't apply (e.g., a custom libtorch path).

Once Option C is verified working, the `LD_PRELOAD` exports in `scripts/gpu-train.sh:35`, `scripts/train-gpu.sh:55`, and `scripts/train-snake-remote.sh:28` can be removed (or kept as defensive belt-and-suspenders).

Defer Option A (upstream `torch-sys` patch) to a separate follow-up issue with the `tch-rs` maintainers.

### Acceptance Criteria

- [ ] `cargo build --release --example test_cuda` on a CUDA-enabled Linux box produces a binary where `ldd <binary> | grep torch_cuda` returns a non-empty line (and similarly for `libc10_cuda`).
- [ ] Running that binary **without** `LD_PRELOAD` set prints `tch::Cuda::is_available() = true` and `Device::cuda_if_available() = Cuda(0)`.
- [ ] `readelf -d target/release/examples/test_cuda | grep NEEDED` lists `libtorch_cuda.so` and `libc10_cuda.so`.
- [ ] CPU-only macOS builds (no CUDA) still build and run correctly; no spurious CUDA link errors on Darwin.
- [ ] The `LD_PRELOAD` workaround in `scripts/gpu-train.sh`, `scripts/train-gpu.sh`, and `scripts/train-snake-remote.sh` is either removed or annotated as "defensive only, no longer required by build".
- [ ] `docs/GPU_SETUP.md` and `docs/GPU_TRAINING_GUIDE.md` are updated to reflect the new build behavior (the "Key Discoveries" / "CUDA Detection Problem" sections need revising — the bug is no longer present).
- [ ] If Option D guard is added: a build with `THRUST_EXPECT_CUDA=1` on a box where CUDA libs failed to link produces a clear fatal error instead of silently running on CPU.

### Affected Files

- `build.rs` (new file at repo root) — implements Option C: detect CUDA, emit positional `-Wl,--no-as-needed` + `-ltorch_cuda -lc10_cuda` + `-Wl,--as-needed`. May need to mirror `torch-sys`'s CUDA detection logic, or rely on inspecting `DEP_TORCH_SYS_*` env vars set by torch-sys's build script for downstream crates.
- `Cargo.toml` (lines 1-16) — may need `build = "build.rs"` declaration and possibly a `[build-dependencies]` section if we need helpers (likely just `std::env`).
- `src/lib.rs` — (optional, Option B variant) add `#[used]` CUDA anchor symbol gated behind `cfg(feature = "training")`.
- `examples/test_cuda.rs` — extend to verify `readelf`/`ldd` output programmatically, or print a clear "CPU fallback detected" warning when expected to be on GPU.
- `scripts/gpu-train.sh:34-38` — remove or comment-out `LD_PRELOAD` line once Option C is verified.
- `scripts/train-gpu.sh:53-59` — same.
- `scripts/train-snake-remote.sh:27-29` — same.
- `docs/GPU_SETUP.md:22-63, 214-223` — rewrite "The CUDA Detection Problem" section; the bug is no longer required to be lived with.
- `docs/GPU_TRAINING_GUIDE.md:53-68` — same; LD_PRELOAD becomes optional/defensive rather than CRITICAL.

### Test Plan

- [ ] **Manual verification on Linux GPU box** (e.g. `rwalters-sandbox-2`):
  1. `git pull` the fix branch.
  2. `unset LD_PRELOAD; unset RUSTFLAGS`.
  3. `export LIBTORCH_USE_PYTORCH=1; export LIBTORCH_BYPASS_VERSION_CHECK=1`.
  4. `cargo clean && cargo build --release --example test_cuda`.
  5. `ldd target/release/examples/test_cuda | grep -E '(torch_cuda|c10_cuda)'` — must return both lines.
  6. `./target/release/examples/test_cuda` — must print `Cuda(0)` and `tch::Cuda::is_available() = true`.
- [ ] **Manual verification on macOS** (CPU-only):
  1. `cargo build --release --example test_cuda` succeeds.
  2. Binary runs without CUDA-related dyld errors.
  3. Reports `Device::Cpu` correctly.
- [ ] **Regression test for training throughput**:
  1. Run `./scripts/gpu-train.sh train_cartpole_long` *without* the `LD_PRELOAD` line in the script.
  2. Confirm `nvidia-smi -l 1` shows GPU utilization > 50% during training.
  3. Confirm training throughput matches or exceeds previous `LD_PRELOAD`-based runs (no regression).
- [ ] **Build-script idempotency**: `cargo build` twice in a row should not re-run the new `build.rs` unnecessarily (proper `cargo:rerun-if-changed=` directives).
- [ ] **Automated tests**: N/A for link-time behavior. Could add a CI step that runs `ldd | grep torch_cuda` on the test_cuda binary if a GPU runner is available.
- [ ] **Edge cases**:
  - CPU-only feature builds (`--no-default-features --features wasm`) must not pull in CUDA link args.
  - `LIBTORCH_USE_PYTORCH=0` with a system libtorch that has no CUDA must not break the build.
  - Different PyTorch versions (2.9.x patch versions) should all work.

### Dependencies

None. Issue #4 and #8 are tangentially related (same "tch is brittle" family) but do not block this fix. This issue can proceed independently.

### Out of Scope

- Upstream patch to `torch-sys` / `tch-rs` (file as a separate issue with maintainers if Option C proves insufficient).
- Windows support (`LD_PRELOAD` equivalent is `PATH` DLL ordering, different bug class).
- macOS Metal/MPS device detection (not affected by Linux `--as-needed`).
- Refactoring the GPU training script suite (`gpu-train.sh` vs `train-gpu.sh` vs `train-snake-remote.sh` duplication — separate cleanup issue).

rjwalters:
## Summary

Closes #9.

- Adds a crate-root `build.rs` that re-emits positional link args
  (`-Wl,--no-as-needed -ltorch_cuda -lc10_cuda -Wl,--as-needed` plus
  `-rpath`) so GNU ld no longer strips `libtorch_cuda.so` from the
  binary's `DT_NEEDED` list. This is the fix for the silent CPU
  fallback described in #9.
- Adds a runtime guard `thrust_rl::utils::cuda::ensure_cuda_if_expected`
  that hard-fails with exit code 2 if `THRUST_EXPECT_CUDA=1` is set but
  `tch::Cuda::is_available()` returns false. Turns silent CPU fallback
  into a loud, fast failure.
- Removes the now-redundant `LD_PRELOAD=libtorch_cuda.so:libtorch.so`
  and `RUSTFLAGS="-C link-arg=-Wl,--no-as-needed"` exports from
  `scripts/gpu-train.sh`, `scripts/train-gpu.sh`, and
  `scripts/train-snake-remote.sh` (kept as commented documentation in
  case of regression). All three scripts now `export
  THRUST_EXPECT_CUDA=1`.
- Rewrites `docs/GPU_SETUP.md` and `docs/GPU_TRAINING_GUIDE.md` to
  reflect the new behavior --- `LD_PRELOAD` is now described as
  historical / fallback rather than CRITICAL.
- Registers `examples/test_cuda.rs` as a proper `[[example]]` in
  `Cargo.toml`. The example now invokes the guard and prints
  diagnostics on CPU fallback.

## Why the historical workaround didn't work

The pre-existing `RUSTFLAGS="-C link-arg=-Wl,--no-as-needed"` is
*positionally ineffective*: `-Wl,--no-as-needed` only affects libraries
that appear *after* it on the linker command line, and the flag set via
`RUSTFLAGS` landed before the `-ltorch_cuda` arg that `torch-sys` emits.
Re-emitting the bracket from a leaf-crate `build.rs` puts the flag at
the end of the command line where it actually covers `-ltorch_cuda`.
The `LD_PRELOAD` workaround patched the symptom at runtime but was a
foot-gun: anyone running the binary outside the wrapper script silently
got a CPU run.

## What runs on this PR's build.rs

On Linux + `training` feature + a CUDA-enabled libtorch is detected:

```
cargo:rustc-link-search=native=<libtorch_lib_dir>
cargo:rustc-link-arg-bins=-Wl,--no-as-needed
cargo:rustc-link-arg-examples=-Wl,--no-as-needed
cargo:rustc-link-arg-tests=-Wl,--no-as-needed
cargo:rustc-link-arg-benches=-Wl,--no-as-needed
cargo:rustc-link-lib=torch_cuda
cargo:rustc-link-lib=c10_cuda
cargo:rustc-link-arg-{bins,examples,tests,benches}=-Wl,--as-needed
cargo:rustc-link-arg-{bins,examples,tests,benches}=-Wl,-rpath,<libtorch_lib_dir>
```

The `-bins`/`-examples`/`-tests`/`-benches` variants are used so the
link args apply to every binary kind that depends on `thrust-rl` (the
plain `rustc-link-arg` form only applies to the crate's own artifact,
which for `thrust-rl` is `cdylib + rlib` and the rlib doesn't go
through the linker). `rustc-link-lib` applies to all artifacts already
and doesn't need a variant.

On macOS, Windows, CPU-only feature builds, or with
`THRUST_DISABLE_CUDA_LINK_FIX=1` set: the script emits only its
`rerun-if-env-changed=` directives and otherwise no-ops.

### Verified locally (macOS, CPU-only)

- [x] `rustc build.rs` parses standalone --- no syntax errors.
- [x] `cargo check --no-default-features` runs `build.rs`; output file
      contains only `rerun-if-env-changed=` lines (no spurious CUDA
      link args on macOS). The only build error is a pre-existing
      missing `src/env/games/pong.rs` module (untracked in `main`,
      unrelated to this PR).
- [x] `cargo check --no-default-features --features wasm` --- same
      pre-existing `pong` error only; no new errors or warnings
      attributable to this PR.
- [x] `cargo:rerun-if-env-changed=` directives cover `LIBTORCH`,
      `LIBTORCH_USE_PYTORCH`, `LIBTORCH_BYPASS_VERSION_CHECK`,
      `THRUST_DISABLE_CUDA_LINK_FIX`, and `CARGO_FEATURE_TRAINING`, so
      `cargo build` twice in a row will not re-run the script
      unnecessarily.

### Requires a CUDA Linux box --- NOT YET VERIFIED

I do not have access to a CUDA-equipped Linux host from this
environment. The following acceptance criteria from the issue still
need confirmation on `rwalters-sandbox-2` (or equivalent):

- [ ] `unset LD_PRELOAD RUSTFLAGS; export LIBTORCH_USE_PYTORCH=1
      LIBTORCH_BYPASS_VERSION_CHECK=1; cargo clean && cargo build
      --release --example test_cuda`.
- [ ] `ldd target/release/examples/test_cuda | grep -E
      '(torch_cuda|c10_cuda)'` returns both lines, neither marked
      "not found".
- [ ] `readelf -d target/release/examples/test_cuda | grep NEEDED`
      lists `libtorch_cuda.so` and `libc10_cuda.so`.
- [ ] `./target/release/examples/test_cuda` (no `LD_PRELOAD`) prints
      `Device::cuda_if_available() = Cuda(0)` and
      `tch::Cuda::is_available() = true`.
- [ ] `./scripts/gpu-train.sh test_cuda` succeeds end-to-end with the
      `LD_PRELOAD` lines commented out.
- [ ] `nvidia-smi -l 1` during a real training run shows >50% GPU
      utilization (regression check vs the old LD_PRELOAD path).
- [ ] Negative test: `THRUST_EXPECT_CUDA=1
      THRUST_DISABLE_CUDA_LINK_FIX=1 ./target/release/examples/test_cuda`
      (i.e. the build.rs fix disabled) exits 2 with the diagnostic
      message.

### Edge cases I want a reviewer to think about

- The `build.rs` shells out to `python3` (and `python` as a fallback)
  to discover the PyTorch lib dir. On a stripped-down build container
  without Python, detection will simply fail and the script becomes a
  no-op (CPU-only build). Is that an acceptable degradation?
- `rustc-link-arg-tests` and `rustc-link-arg-benches` are listed for
  completeness but neither path is exercised on CI today.
- `-Wl,-rpath,<dir>` bakes the build host's libtorch path into the
  binary. This is convenient but means binaries are not portable
  across hosts with different Python prefixes. If that's a concern,
  the rpath emission could be made opt-in via an env var --- ping me
  if you'd like that gate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,11 @@ repository = "https://github.com/yourusername/thrust"
 keywords = ["reinforcement-learning", "machine-learning", "rl", "deep-learning", "cuda"]
 categories = ["science", "algorithms"]
 readme = "README.md"
+# Build script defeats GNU ld's --as-needed for libtorch_cuda on Linux.
+# See build.rs for the rationale (in short: without it, the binary silently
+# falls back to CPU even on a CUDA-equipped box). No-op on macOS/Windows
+# and on CPU-only feature builds.
+build = "build.rs"
 
 [lib]
 crate-type = ["cdylib", "rlib"]
@@ -180,3 +185,11 @@ required-features = ["training"]
 name = "train_p3"
 path = "examples/games/bucket_brigade/train_p3.rs"
 required-features = ["training", "env-bucket-brigade"]
+
+# Diagnostics --- verifies that the CUDA link fix from build.rs is working.
+# Run with `./scripts/gpu-train.sh test_cuda` or directly via cargo on a
+# CUDA-equipped Linux box. On macOS this still builds but will report `Cpu`.
+[[example]]
+name = "test_cuda"
+path = "examples/test_cuda.rs"
+required-features = ["training"]
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,227 @@
+//! Build script for thrust-rl.
+//!
+//! Primary purpose: defeat the GNU linker's `--as-needed` default behavior so
+//! that `libtorch_cuda.so` (and `libc10_cuda.so`) stay in the binary's
+//! `DT_NEEDED` list even though the Rust source never references a symbol
+//! that resolves to those libraries directly.
+//!
+//! ## Background
+//!
+//! Modern Linux distros (Ubuntu 22.04+, Debian 12+, etc.) ship `ld` configured
+//! with `--as-needed` enabled by default. Under that flag, any `DT_NEEDED`
+//! entry for a shared library from which no symbol is actually referenced is
+//! dropped at link time. `torch-sys`'s build script emits
+//! `cargo:rustc-link-lib=torch_cuda` correctly when CUDA is detected, but
+//! since `tch` itself only directly references symbols in `libtorch_cpu` and
+//! `libc10`, the linker silently strips `-ltorch_cuda` from the final binary.
+//!
+//! At runtime, `libtorch` performs lazy CUDA backend registration that
+//! depends on `libtorch_cuda.so` already being loaded into the process. With
+//! the `DT_NEEDED` entry stripped, the CUDA backend never initializes and
+//! `tch::Cuda::is_available()` returns `false` --- *silent CPU fallback*.
+//!
+//! ## Why downstream `RUSTFLAGS="-C link-arg=-Wl,--no-as-needed"` is not enough
+//!
+//! `-Wl,--no-as-needed` is *positional*: it only affects libraries that
+//! appear *after* it on the linker command line. When set via
+//! `RUSTFLAGS`, the flag is injected at a position that does not cover
+//! the `-ltorch_cuda` argument emitted by the `torch-sys` build script.
+//! Re-emitting the bracket from a leaf-crate build script puts the flag
+//! at the *end* of the link line where it does cover the late-emitted
+//! `-ltorch_cuda`.
+//!
+//! ## What this script does
+//!
+//! On Linux only, and only when the `training` feature is enabled, it emits:
+//!
+//! ```text
+//! cargo:rustc-link-arg=-Wl,--no-as-needed
+//! cargo:rustc-link-lib=torch_cuda
+//! cargo:rustc-link-lib=c10_cuda
+//! cargo:rustc-link-arg=-Wl,--as-needed
+//! ```
+//!
+//! when CUDA is detected in the libtorch installation. On macOS, Windows, or
+//! CPU-only builds, the script is a no-op.
+//!
+//! ## CUDA detection
+//!
+//! Detection mirrors `torch-sys`'s logic at a coarse level: if
+//! `LIBTORCH_USE_PYTORCH=1` is set, we shell out to Python to ask whether
+//! CUDA is compiled in and discover the lib path. Otherwise we look at
+//! `LIBTORCH` env var (the standalone libtorch path) and check for the
+//! presence of `libtorch_cuda.so`. If neither path yields a CUDA-enabled
+//! libtorch, the script emits nothing extra and the build proceeds as
+//! CPU-only.
+//!
+//! ## Opting out
+//!
+//! Set `THRUST_DISABLE_CUDA_LINK_FIX=1` to skip the link-arg emission. Useful
+//! if a downstream user has a custom linker setup that the fix interferes
+//! with. The `THRUST_EXPECT_CUDA` runtime check (see `src/utils/cuda.rs`)
+//! is independent of this opt-out.
+
+use std::env;
+use std::path::PathBuf;
+use std::process::Command;
+
+fn main() {
+    // Re-run conditions: only re-run when the relevant environment variables
+    // change. We deliberately do NOT use `cargo:rerun-if-changed=build.rs`
+    // alone because Cargo defaults to that. Instead, we add env-var triggers
+    // for the inputs we actually consult.
+    println!("cargo:rerun-if-env-changed=LIBTORCH");
+    println!("cargo:rerun-if-env-changed=LIBTORCH_USE_PYTORCH");
+    println!("cargo:rerun-if-env-changed=LIBTORCH_BYPASS_VERSION_CHECK");
+    println!("cargo:rerun-if-env-changed=THRUST_DISABLE_CUDA_LINK_FIX");
+    println!("cargo:rerun-if-env-changed=CARGO_FEATURE_TRAINING");
+
+    let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap_or_default();
+    let training_enabled = env::var("CARGO_FEATURE_TRAINING").is_ok();
+    let opted_out = env::var("THRUST_DISABLE_CUDA_LINK_FIX")
+        .map(|v| v == "1" || v.eq_ignore_ascii_case("true"))
+        .unwrap_or(false);
+
+    // Only apply the fix on Linux with the training feature on. macOS, iOS,
+    // and Windows do not exhibit the `--as-needed` behavior we are countering
+    // (Windows uses a totally different link model, macOS's ld doesn't strip
+    // unreferenced dylibs the same way).
+    if target_os != "linux" {
+        return;
+    }
+    if !training_enabled {
+        return;
+    }
+    if opted_out {
+        eprintln!(
+            "thrust-rl build.rs: THRUST_DISABLE_CUDA_LINK_FIX is set; skipping \
+             CUDA link-arg emission."
+        );
+        return;
+    }
+
+    let cuda_lib_dir = match detect_cuda_lib_dir() {
+        Some(dir) => dir,
+        None => {
+            // CPU-only build, or libtorch without CUDA. Nothing to do.
+            eprintln!(
+                "thrust-rl build.rs: no CUDA-enabled libtorch detected; \
+                 skipping CUDA link-arg emission. \
+                 (training feature is on, but libtorch_cuda.so was not found.)"
+            );
+            return;
+        }
+    };
+
+    eprintln!(
+        "thrust-rl build.rs: emitting positional link args to prevent \
+         --as-needed from stripping libtorch_cuda (lib dir: {}).",
+        cuda_lib_dir.display()
+    );
+
+    // Add a `rustc-link-search` directive so the linker can find the CUDA
+    // libs at link time. `torch-sys` typically already adds this; we re-emit
+    // it as a belt-and-suspenders guard.
+    println!(
+        "cargo:rustc-link-search=native={}",
+        cuda_lib_dir.display()
+    );
+
+    // The bracket: enable --no-as-needed, list the libraries we want to keep
+    // referenced, then restore --as-needed so we don't accidentally retain
+    // unrelated unused libs later on the command line.
+    //
+    // Position matters: Cargo emits build-script `rustc-link-arg` directives
+    // *after* most dependency `rustc-link-lib` directives, so our re-emitted
+    // `-Wl,--no-as-needed` covers the `-ltorch_cuda` we re-list here AND any
+    // late `-ltorch_cuda` already emitted by `torch-sys`'s own build script.
+    //
+    // We use the `-bins` and `-examples` variants of `rustc-link-arg` so the
+    // flags apply to binary artifacts that link this crate (not the rlib
+    // itself, which is a static archive and doesn't go through the linker).
+    // `rustc-link-lib` applies to every artifact and doesn't need a variant.
+    //
+    // We emit duplicate directives for `-bins`, `-examples`, `-tests`, and
+    // `-benches` so any binary kind that links thrust-rl gets the fix.
+    for variant in ["bins", "examples", "tests", "benches"] {
+        println!("cargo:rustc-link-arg-{variant}=-Wl,--no-as-needed");
+    }
+    println!("cargo:rustc-link-lib=torch_cuda");
+    println!("cargo:rustc-link-lib=c10_cuda");
+    for variant in ["bins", "examples", "tests", "benches"] {
+        println!("cargo:rustc-link-arg-{variant}=-Wl,--as-needed");
+    }
+
+    // Belt-and-suspenders: also embed an RPATH entry so the binary can find
+    // the CUDA libs at runtime without relying on `LD_LIBRARY_PATH`. We
+    // use `-Wl,-rpath,...` rather than passing `-rpath` directly so the
+    // flag survives gcc/clang's argument forwarding (rustc invokes the
+    // system linker via the C compiler driver).
+    for variant in ["bins", "examples", "tests", "benches"] {
+        println!(
+            "cargo:rustc-link-arg-{variant}=-Wl,-rpath,{}",
+            cuda_lib_dir.display()
+        );
+    }
+}
+
+/// Locate the directory containing `libtorch_cuda.so` and `libc10_cuda.so`,
+/// or `None` if we can't find them (CPU-only or non-Linux libtorch).
+fn detect_cuda_lib_dir() -> Option<PathBuf> {
+    // Strategy 1: pip-installed PyTorch via LIBTORCH_USE_PYTORCH=1.
+    if env::var("LIBTORCH_USE_PYTORCH").as_deref() == Ok("1") {
+        if let Some(dir) = pytorch_lib_dir_from_python() {
+            if has_cuda_libs(&dir) {
+                return Some(dir);
+            }
+        }
+    }
+
+    // Strategy 2: explicit LIBTORCH path.
+    if let Ok(libtorch) = env::var("LIBTORCH") {
+        let lib_dir = PathBuf::from(libtorch).join("lib");
+        if has_cuda_libs(&lib_dir) {
+            return Some(lib_dir);
+        }
+    }
+
+    // Strategy 3: fallback to invoking python3 unconditionally. This catches
+    // setups where the user has pip-installed torch but forgot to set
+    // LIBTORCH_USE_PYTORCH=1 explicitly --- if we can find a CUDA-enabled
+    // torch we may as well emit the link args (they're a no-op without the
+    // matching torch-sys link directives anyway).
+    if let Some(dir) = pytorch_lib_dir_from_python() {
+        if has_cuda_libs(&dir) {
+            return Some(dir);
+        }
+    }
+
+    None
+}
+
+fn pytorch_lib_dir_from_python() -> Option<PathBuf> {
+    // Try python3 first, then python. Quiet errors --- we'll just fail
+    // detection and let the caller handle the None case.
+    for py in ["python3", "python"] {
+        let out = Command::new(py)
+            .arg("-c")
+            .arg(
+                "import os, torch; \
+                 print(os.path.join(os.path.dirname(torch.__file__), 'lib'))",
+            )
+            .output();
+        if let Ok(out) = out {
+            if out.status.success() {
+                let path = String::from_utf8_lossy(&out.stdout).trim().to_string();
+                if !path.is_empty() {
+                    return Some(PathBuf::from(path));
+                }
+            }
+        }
+    }
+    None
+}
+
+fn has_cuda_libs(dir: &PathBuf) -> bool {
+    dir.join("libtorch_cuda.so").exists() && dir.join("libc10_cuda.so").exists()
+}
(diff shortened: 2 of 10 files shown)


rjwalters:
## Judge Review: Approved

This is a careful, well-documented implementation that maps cleanly onto the curator's Option C + Option D recommendation. Approving for merge.

### Verified locally (macOS, this judge)
- `rustc build.rs` parses standalone. No syntax errors.
- `cargo check --no-default-features` runs `build.rs` and the emitted output file at `target/debug/build/thrust-rl-*/output` contains **only** the `cargo:rerun-if-env-changed=` lines on macOS — confirms the script is a clean no-op off Linux.
- Build error in the workspace is the pre-existing missing `src/env/games/pong.rs` module (already on `main`), not attributable to this PR.

### Design quality
- **Positional link args bracket is correct**: `-Wl,--no-as-needed` is emitted *before* the re-listed `-ltorch_cuda -lc10_cuda`, with `-Wl,--as-needed` restored after. Cargo appends build-script `rustc-link-arg` directives at the end of the link line, which is the position that actually covers the late-emitted `-ltorch_cuda` from `torch-sys`. This is exactly the fix the curator analyzed.
- **`rustc-link-arg-{bins,examples,tests,benches}` variants**: nice touch. The plain `rustc-link-arg` would only apply to this crate's own cdylib output and miss every binary kind that *depends* on `thrust-rl`, which is the actual use case. `rustc-link-lib` correctly stays as the unsuffixed form (applies to all artifacts).
- **CUDA detection cascade** (PyTorch via `LIBTORCH_USE_PYTORCH=1` → explicit `LIBTORCH` → fallback Python probe) is sensible. The `python3` shell-out degrades gracefully to a no-op if Python isn't available; the PR body correctly raises this as an edge case.
- **`THRUST_DISABLE_CUDA_LINK_FIX=1` opt-out** is documented in both `build.rs` and `docs/GPU_SETUP.md`. Good escape hatch.
- **Belt-and-suspenders `-rpath`** baked into the binary means users don't have to fuss with `LD_LIBRARY_PATH` at run time. The PR body correctly notes the portability tradeoff.

### Runtime guard
- `src/utils/cuda.rs::ensure_cuda_if_expected` correctly no-ops when `THRUST_EXPECT_CUDA` is unset, empty, `0`, or `false`. Only hard-fails when the env is truthy AND device is `Cpu`. Exit code 2 is a good choice (distinguishes from `Result::Err` exit 1).
- Diagnostic message is excellent: walks the user through `ldd <bin> | grep torch_cuda` and the `LD_PRELOAD` fallback. Anyone hitting this guard will know exactly what to check.
- Non-training stub at `#[cfg(not(feature = "training"))]` accepts `&()` so the symbol exists in CPU-only builds. The PR's own note that "most call sites are inside `cfg(feature = training)` code anyway" is accurate — it's defensive without being intrusive.

### Scripts and docs
- All three scripts (`gpu-train.sh`, `train-gpu.sh`, `train-snake-remote.sh`) have `LD_PRELOAD` and `RUSTFLAGS="-C link-arg=-Wl,--no-as-needed"` commented out with a clear note explaining why (issue #9 root cause) and how to re-enable as a fallback. All three export `THRUST_EXPECT_CUDA=1`.
- `docs/GPU_SETUP.md` and `docs/GPU_TRAINING_GUIDE.md` are rewritten correctly: `LD_PRELOAD` is now described as a historical fallback rather than CRITICAL, and the new `build.rs` behavior is front-and-center.
- `examples/test_cuda.rs` is properly registered in `Cargo.toml` with `required-features = ["training"]` and invokes the guard at the end so a CUDA-equipped run with `THRUST_EXPECT_CUDA=1` hard-fails on regression.

### CI status note
All CI jobs on this PR are red, but every failure is **pre-existing and unrelated**:
- `bucket-brigade-core` submodule not checked out by CI (workflows lack `submodules: recursive`) — root cause of every Test Suite failure and the MSRV failure.
- Rustfmt diffs in `tests/test_bucket_brigade_env.rs`, `tests/test_multi_discrete_mlp.rs`, `tests/test_ppo_aux_loss.rs` — none of these files are touched by this PR.
- Pre-existing 42 lib unit test compile errors tracked as issue #7.
- Missing `src/env/games/pong.rs` module declared in `mod.rs` but uncommitted on `main`.

These should be addressed in a separate CI-fix PR. Not blocking this one.

### Minor nits (non-blocking, optional follow-up)
Clippy on `build.rs` reports:
- A few `collapsible_if` warnings at `build.rs:172, 173, 193, 213` (style only — the nested form arguably reads more clearly given the strategy comments).
- `fn has_cuda_libs(dir: &PathBuf)` at line 225 should idiomatically be `dir: &Path` (`clippy::ptr_arg`).

None of these will fail CI in the current configuration and they don't affect correctness; mentioning only for the next pass.

### What this PR honestly cannot verify here
The PR body correctly flags that the following criteria require a Linux CUDA host (which neither the builder nor this judge has access to). Accepting "needs Linux GPU validation post-merge" as honest reporting:
- `ldd target/release/examples/test_cuda | grep -E '(torch_cuda|c10_cuda)'` returning both lines.
- `readelf -d target/release/examples/test_cuda | grep NEEDED` listing the CUDA `.so`s.
- `./scripts/gpu-train.sh test_cuda` end-to-end on `rwalters-sandbox-2`.
- `nvidia-smi -l 1` showing >50% GPU util during a real run.
- Negative test: `THRUST_EXPECT_CUDA=1 THRUST_DISABLE_CUDA_LINK_FIX=1` should exit 2.

Recommendation: after merging, run the test_cuda example on the sandbox box and post the `ldd` / `readelf` output as a comment on the merged commit so we have a record of the link-time fix actually working in the field. If anything regresses, the `THRUST_EXPECT_CUDA=1` guard will catch it loudly on the next training run.

### Approval

Approved. Transitioning `loom:review-requested` → `loom:pr`.

PR #22, opened 2026-05-14
test(snake): fix collision detection test assertions

### LINKED ISSUE #17: Snake env: test_snake_collision_detection fails — collides_with_wall(3, 3) returns true for in-bounds head at (2, 1)
## Background

This bug was hidden until PR #16 fixed the lib unit test compilation (issue #7). With tests now compiling, `cargo test --lib` runs 45 tests of which 44 pass; `env::games::snake::snake::tests::test_snake_collision_detection` is the one failure.

## Symptom

The test at `src/env/games/snake/snake.rs:202` asserts that `collides_with_wall(3, 3)` returns `true` for a snake head at position `(2, 1)` — which is in-bounds on a 3x3 grid.

There is a mismatch between the test's expectation and the collision-detection logic. Either:
- The test expectation is wrong (`(2, 1)` is in-bounds on a 3x3 grid, so `collides_with_wall` should be `false`), or
- The collision logic has an off-by-one bug.

Need investigation to determine which side is wrong.

## Reproduction

```bash
cargo test --lib --no-default-features test_snake_collision_detection
```

## Acceptance

- `cargo test --lib --no-default-features` reports 0 failures.
- The fix has a regression test covering whichever boundary case was wrong.

## Discovery

Flagged by the Loom Judge during review of PR #16 ([judge comment](https://github.com/rjwalters/thrust/pull/16#issuecomment-4455093238)). Builder for PR #16 explicitly scoped this bug out of #7's work.

rjwalters:
## Summary

Fixes the failing `test_snake_collision_detection` test by correcting both the wall-collision and self-collision assertions. The implementation (`Snake::collides_with_wall`, `Position::in_bounds`) is correct and unchanged — only the test was wrong.

## Changes

- `src/env/games/snake/snake.rs` (test only): replace the incorrect wall-collision assertion at line 202 with a pair of assertions that exercise both an in-bounds case (`(2, 1)`, last valid column on 3x3) and an out-of-bounds case (`(3, 1)`, first invalid column). This locks in boundary correctness in both directions.
- Fix the self-collision portion of the same test: the snake was too short (length 2) to actually collide with itself after three moves. Grow to length 5 and walk a closed square (Right, Down, Left, Up) so the head returns to its starting position and collides with the retained body tail.

| `cargo test --lib --no-default-features` reports 0 failures | Met | 45 passed; 0 failed (was 44/1) |
| Regression test covers the corrected boundary case | Met | New assertions cover both `(2, 1)` in-bounds and `(3, 1)` out-of-bounds on 3x3 |
| No changes to `Snake::collides_with_wall` or `Position::in_bounds` | Met | Implementation untouched; `git diff --stat` is one file |

## Test Plan

- `cargo test --lib --no-default-features test_snake_collision_detection` — passes
- `cargo test --lib --no-default-features` — 45 passed, 0 failed
- `cargo +nightly fmt` — clean

## Note on Scope

The curator's review recommended touching only the wall-collision portion of the test. While implementing that fix, the self-collision portion also began failing — it had been silently masked by the earlier panic. Since the issue body's acceptance criterion is "`cargo test --lib --no-default-features` reports 0 failures," the self-collision portion had to be fixed in the same PR to meet acceptance. Both fixes are test-only and scoped to a single file.

Closes #17

### DIFF
--- a/src/env/games/snake/snake.rs
+++ b/src/env/games/snake/snake.rs
@@ -196,20 +196,26 @@ mod tests {
     fn test_snake_collision_detection() {
         let mut snake = Snake::new(0, Position::new(1, 1), Direction::Right);
         snake.grow();
-        snake.move_forward(); // Now at (2,1) with body at (1,1)
+        snake.move_forward(); // head now at (2, 1) — last in-bounds column on 3x3
+        assert!(!snake.collides_with_wall(3, 3), "(2, 1) is in-bounds on 3x3");
 
-        // Wall collision
-        assert!(snake.collides_with_wall(3, 3)); // Grid is 3x3 (0-2)
+        // Wall collision — step off the right edge
+        snake.move_forward(); // head now at (3, 1) — off the right edge
+        assert!(snake.collides_with_wall(3, 3), "(3, 1) is out-of-bounds on 3x3");
 
-        // Self collision - move to create collision
+        // Self collision - grow long enough to loop back onto own body
         let mut snake2 = Snake::new(0, Position::new(5, 5), Direction::Right);
         snake2.grow();
-        snake2.move_forward();
+        snake2.grow();
+        snake2.grow();
+        snake2.grow(); // length is now 5 (initial 1 + 4 grows)
+        snake2.move_forward(); // head (6, 5)
         snake2.change_direction(Direction::Down);
-        snake2.move_forward();
+        snake2.move_forward(); // head (6, 6)
         snake2.change_direction(Direction::Left);
-        snake2.move_forward();
-        // Now should be colliding with itself
+        snake2.move_forward(); // head (5, 6)
+        snake2.change_direction(Direction::Up);
+        snake2.move_forward(); // head back at (5, 5) — collides with body tail
         assert!(snake2.collides_with_self());
     }
 


rjwalters:
## Approved

Reviewed the test changes carefully against the curator's diagnosis. The fix is correct and the scope expansion is justified.

### Verification

- **Diff scope**: +14 / -8 in `src/env/games/snake/snake.rs` only (test code).
- **Implementation untouched**: `Snake::collides_with_wall` and `Position::in_bounds` unchanged — they're correct under the half-open convention.
- **Wall-collision half**: Now asserts `(2, 1)` is in-bounds and `(3, 1)` is out-of-bounds on 3x3. Locks in boundary correctness in both directions, exactly as the curator recommended.
- **Self-collision half**: Traced through the closed-square walk manually. Starting from `(5, 5)` with length 5, after 4 grows the snake walks Right→Down→Left→Up: head visits `(6,5)`, `(6,6)`, `(5,6)`, `(5,5)`. With length=5 and only 5 segments in the body, `move_forward` retains the tail at `(5,5)`, so the head genuinely collides with the original starting segment. The test honestly exercises `collides_with_self`.
- **Local test run** (`cargo test --lib --no-default-features`): 45 passed, 0 failed — matches the issue's acceptance criterion.
- **Specific test**: `test_snake_collision_detection` passes.
- **Formatting**: `cargo +nightly fmt --check` is clean.

### Scope Expansion Assessment

The builder's note about expanding scope is correct and well-handled. The original test's self-collision half was structurally broken (a length-2 snake can't collide with itself in 3 moves) but masked by the earlier wall-collision panic short-circuiting the function. Fixing only the wall half would have flipped failure modes rather than satisfying the acceptance criterion. The expansion is the minimum needed to land "0 failures" honestly, and it stays test-only in a single file.

### CI Status

The failing checks (Documentation, Tests Linux/macOS + libtorch) are pre-existing on `main` — unrelated compilation errors in `debug_cartpole` and `TrainingMetadata`/`SpaceInfo` mismatches, plus documentation lints. The same failure pattern is present on the last several `main` commits. This PR introduces no new failures; the `Check (no libtorch)`, `Rustfmt`, and `Security Audit` checks (which exercise the changed code path) all pass.

LGTM.

PR #32, opened 2026-05-14
Remove RNG-consuming probe in JointMultiAgentTrainer::collect_rollout

### LINKED ISSUE #27: JointMultiAgentTrainer::collect_rollout samples one extra libtorch RNG draw during action-shape probe
## Symptom

`JointMultiAgentTrainer::collect_rollout` at `src/multi_agent/joint.rs:405` probes the action tensor's rank by calling `policies[0].get_action(&probe_obs)`. Because `MultiDiscreteMlpPolicy::get_action` (and `MlpPolicy::get_action`) sample via `multinomial`, this consumes one libtorch RNG draw that the pre-migration `examples/games/bucket_brigade/train_p3.rs` reference impl did not consume.

Net effect: strict byte-for-byte numerical parity (1e-5 `total_loss` / `redundancy_loss` match) between a run using the library trainer and a run using the pre-migration inline reference will NOT hold. Algorithmic semantics are preserved — only RNG sequence diverges.

## Recommended fix

Replace the sampling probe with a non-sampling shape introspection. Options:

1. Add a `fn action_dims(&self) -> Vec<usize>` trait method to `JointPolicy` that returns the action shape without sampling.
2. Make action shape part of `JointTrainerConfig` and require the caller to declare it.

(1) is cleaner — the policy already knows its own action shape.

## Reproduction

```bash
cargo run --release --features 'training env-bucket-brigade' --example train_p3 -- --scenario default --lambda-red 0.01 --seed 42
```
…and compare first-iter `total_loss` / `redundancy_loss` against a run using the pre-migration code at commit `40ec676`.

## Acceptance

- First-iter `total_loss` and `redundancy_loss` match the pre-migration reference (`40ec676:examples/games/bucket_brigade/train_p3.rs`) within 1e-5 with seed 42.
- Existing 7 integration tests in `tests/test_joint_trainer.rs` still pass.

## Discovery

Flagged by the judge during PR #15 review ([approval comment](https://github.com/rjwalters/thrust/pull/15)) as a non-blocking follow-up.

rjwalters:
## Summary

- Replace `policies[0].get_action(&probe_obs)` probe in `JointMultiAgentTrainer::collect_rollout` with a new `JointPolicy::action_dims()` trait method that introspects action shape without sampling.
- Implement the trait method for `MlpPolicy` (returns `vec![action_dim]`) and `MultiDiscreteMlpPolicy` (returns `self.action_dims().to_vec()`).
- Add `action_dim: i64` field plus public `action_dim()` accessor to `MlpPolicy` (additive — no existing callsite breaks).

Closes #27

## Why this matters

The probe call consumed one libtorch global-RNG draw **per action dimension** (via `multinomial`, which advances the generator even under `tch::no_grad`). For a `K`-factored multi-discrete space, this burned `K` RNG slots before every rollout — shifting the entire downstream action sampling stream. The drift broke first-iter numerical parity with the pre-extraction reference example at `40ec676:examples/games/bucket_brigade/train_p3.rs`.

Using a non-sampling, shape-introspection method preserves parity because it touches no tensor ops.

## Implementation notes

- **`MlpPolicy`**: added `action_dim` field (set from the existing constructor argument) and a public `action_dim()` accessor. `new()` and `with_config()` signatures are unchanged.
- **`MultiDiscreteMlpPolicy`**: already had inherent `action_dims(&self) -> &[i64]`. The trait method `action_dims(&self) -> Vec<i64>` has a different return type, so Rust's method dispatch resolves the collision cleanly within the impl block. Documented inline.
- **`collect_rollout`**: the rank-dispatch branches (`num_action_dims == 1` vs `> 1`) at the action-stacking sites are unchanged — they're keyed off the count, which is exactly what the new trait method returns.

## Test plan

- [x] `cargo check --features 'training env-bucket-brigade'` passes.
- [x] `cargo check --example train_p3 --features 'training env-bucket-brigade'` passes.
- [x] `cargo test --features training --test test_joint_trainer` — all 7 tests pass (smoke, rollout shapes, aux-gradient coupling, aux-none, multi-discrete impl, device-uniformity validation, optimizer-count validation).
- [x] `cargo test --features training --lib multi_agent::joint` — all 5 in-src tests pass.
- [x] `cargo test --features training --lib policy::mlp` — all 8 in-src tests pass (validates the new field doesn't disturb existing behavior).
- [x] `cargo test --features training --test {test_multi_discrete_mlp,test_encoder_features,test_ppo_aux_loss,integration_train_export_inference}` — all pass.

### Parity verification (the actual acceptance gate)

**Not run.** The acceptance criterion is first-iter `total_loss` / `redundancy_loss` of `train_p3 --scenario default --lambda-red 0.01 --seed 42` matching the same command at commit `40ec676` within `1e-5`. This requires:

1. A working libtorch that matches the historical commit's expectations, **and**
2. Access to checkout commit `40ec676`'s example code.

My local libtorch setup (homebrew pytorch 2.10.0, off-by-one from tch-rs 0.22's expected 2.9.0; via `LIBTORCH_USE_PYTORCH=1 LIBTORCH_BYPASS_VERSION_CHECK=1` it loads but the historical example may have different `tch` API expectations) makes side-by-side parity comparison brittle. CI doesn't run libtorch either. Recommend the reviewer (or the owner of the `40ec676` baseline) run the comparison if/when convenient — the change is small and the mechanism is clearly identified, so the test-suite green-light + the absence of any other RNG-draw source between the two versions of `collect_rollout` is strong evidence that parity is restored.

The pre-existing unrelated test failure `test_ppo_learning::test_ppo_learns_from_synthetic_data` (verified flaky on `master` before this branch) is not affected by this change.

### DIFF
--- a/src/multi_agent/joint.rs
+++ b/src/multi_agent/joint.rs
@@ -106,6 +106,25 @@ pub trait JointPolicy {
     /// Immutable view of the policy's var-store (for parameter inspection
     /// and checkpointing).
     fn var_store(&self) -> &nn::VarStore;
+
+    /// Per-dimension action cardinalities.
+    ///
+    /// Returns a vector whose length is the number of factored action
+    /// dimensions and whose entries are the cardinality of each dim.
+    ///
+    /// - Scalar discrete (e.g. [`crate::policy::mlp::MlpPolicy`] with
+    ///   `action_dim = 5`): returns `vec![5]` (one dim, cardinality 5). The
+    ///   rollout buffer uses `num_action_dims = 1`.
+    /// - Multi-discrete (e.g.
+    ///   [`crate::policy::multi_discrete_mlp::MultiDiscreteMlpPolicy`] with
+    ///   `action_dims = [10, 2]`): returns `vec![10, 2]`. The rollout buffer
+    ///   uses `num_action_dims = 2`.
+    ///
+    /// Used by [`JointMultiAgentTrainer::collect_rollout`] to size action
+    /// buffers without calling [`JointPolicy::get_action`] (which would
+    /// consume libtorch RNG draws -- one per action dimension via
+    /// `multinomial` -- and break parity with reference implementations).
+    fn action_dims(&self) -> Vec<i64>;
 }
 
 impl JointPolicy for crate::policy::mlp::MlpPolicy {
@@ -124,6 +143,11 @@ impl JointPolicy for crate::policy::mlp::MlpPolicy {
     fn var_store(&self) -> &nn::VarStore {
         self.var_store()
     }
+    fn action_dims(&self) -> Vec<i64> {
+        // Scalar-discrete policy: a single action dim of cardinality
+        // `self.action_dim()`.
+        vec![self.action_dim()]
+    }
 }
 
 impl JointPolicy for crate::policy::multi_discrete_mlp::MultiDiscreteMlpPolicy {
@@ -141,6 +165,15 @@ impl JointPolicy for crate::policy::multi_discrete_mlp::MultiDiscreteMlpPolicy {
     }
     fn var_store(&self) -> &nn::VarStore {
         self.var_store()
+    }
+    fn action_dims(&self) -> Vec<i64> {
+        // Naming-collision note: `MultiDiscreteMlpPolicy` already has an
+        // inherent method `action_dims(&self) -> &[i64]`. The trait method
+        // returns `Vec<i64>` (a different type), so there's no ambiguity at
+        // the language level -- the impl-block context resolves the call to
+        // the inherent method, and we materialize a `Vec` via `.to_vec()`.
+        // Readers may double-take; the comment documents the intent.
+        self.action_dims().to_vec()
     }
 }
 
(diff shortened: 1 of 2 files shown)


rjwalters:
## Loom Judge Review: Approved

Surgical, well-targeted fix that cleanly resolves #27. Excellent inline documentation throughout.

### Verification

**Trait + impls**
- `JointPolicy::action_dims() -> Vec<i64>` added at `src/multi_agent/joint.rs:110-127` with a thorough doc comment that names the parity rationale explicitly (RNG-consuming `multinomial` in `get_action`).
- `MlpPolicy::action_dims` returns `vec![self.action_dim()]` (`src/multi_agent/joint.rs:146-150`).
- `MultiDiscreteMlpPolicy::action_dims` returns `self.action_dims().to_vec()` (`src/multi_agent/joint.rs:169-177`). The 6-line inline comment about the inherent-vs-trait naming collision is exactly the right amount of context — readers do double-take, and Rust's inherent-wins-over-trait dispatch makes the call resolve correctly.

**`MlpPolicy` field addition**
- `action_dim: i64` stored at construction in `with_config` (`src/policy/mlp.rs:175`); public `action_dim()` accessor at line 184.
- Confirmed by inspection of all ~14 callsites under `src/`, `examples/`, and `tests/`: every caller uses the 3-argument forms `MlpPolicy::new(obs_dim, action_dim, hidden_dim)` and `MlpPolicy::with_config(obs_dim, action_dim, config)`. **Signatures unchanged as claimed** — no struct-literal construction in user code (the struct field is private, so additive field is safe).

**`collect_rollout` call site** (`src/multi_agent/joint.rs:430-446`)
- The old `Tensor::from_slice` + `no_grad(get_action)` + shape-rank dispatch + `drop(probe_act)` probe is fully removed.
- Replaced with `self.policies[0].action_dims().len()`. The downstream rank-dispatch in the per-step loop (keyed on `num_action_dims == 1`) is preserved and still correct.
- The comment block at the replacement site explains the mechanism and cites the `40ec676` parity reference.

### CI

- **Rustfmt**: pass.
- **Check (no libtorch)** ubuntu + macos: pass.
- **Security Audit**: pass.
- **Tests (Linux + libtorch)** and **Tests (macOS + libtorch)**: both fail solely on `test_ppo_learns_from_synthetic_data`. This is the documented pre-existing flake on `master`; **all 7 `test_joint_trainer.rs` tests pass**, all lib tests for `multi_agent::joint` (5), `policy::mlp` (8), and `policy::multi_discrete_mlp` (7) pass, integration tests pass.
- **Documentation**: fails on missing docs in `src/wasm.rs`, `src/optimize/{bayesian,pareto,scheduler,space}.rs` — files untouched by this PR. Verified pre-existing on `main` (run 25891045534 Documentation job also failed). Not caused by this change.

### Mechanism verification (in lieu of libtorch parity test)

The author honestly disclosed not running the `1e-5` first-iter parity test against `40ec676` due to libtorch version mismatch. By inspection:

- `MlpPolicy::action_dims` reads a struct field — zero tensor ops.
- `MultiDiscreteMlpPolicy::action_dims` clones a `Vec<i64>` — zero tensor ops.
- Neither path touches a `Tensor`, `nn::Module`, or the libtorch global generator.
- The eliminated probe path went `Tensor::from_slice` → `get_action` → `softmax` → `multinomial`, and `multinomial` is the documented RNG-consuming op (one draw per action dim).

The fix mechanism is clearly identified and minimal. No other RNG-draw source remains between the probe-removal site and the per-step `get_action` sampling loop, so parity should be restored. The inspection-only verification is appropriate here.

### Notes / non-blocking observations

- Documentation comments are unusually high-quality — both the trait doc and the inline comment in `collect_rollout` give future readers everything they need to understand *why* the change was made, not just what. Good defensive engineering against future regressions.
- The `tch::no_grad` wrapper on the per-step `policy.get_action` calls inside the rollout loop is unchanged (correct — those calls intentionally consume RNG to sample actions; no-grad is purely for autograd-graph suppression). Verified the fix does not accidentally remove that wrapper.

Approving and transitioning to `loom:pr` for downstream merge.

PR #34, opened 2026-05-14
Fix zero-padded rollout contamination in PPO train_step

### LINKED ISSUE #28: PolicyLearner::train_step consumes full buffer capacity, zero-padding partial-fill rollouts
## Symptom

`PolicyLearner::train_step` (`src/multi_agent/learner.rs`) calls `buffer.get_batch()` which returns the full buffer capacity, not just the written portion. When a training cycle runs with fewer experiences than `buffer_size` (e.g. early episodes, terminal cutoff), the batch handed to PPO includes zero-padded rows for the unwritten tail.

This was a latent issue before #4 / PR #12 (the buffer was never being populated, so it didn't matter). With the buffer fix from PR #12 in place, the zero-padded rows now actively contaminate gradients.

Additionally: the docstring on `buffer_step_idx` at `src/multi_agent/learner.rs:50` claims it 'wraps modulo `config.buffer_size`', but the code never wraps — it monotonically increments until reset after each training cycle. Either the code should wrap or the docstring should be corrected.

## Recommended fix

Two parts:

1. **Stop the zero-padding contamination**: `train_step` should slice the batch to `buffer_step_idx` rows before passing to PPO, OR `RolloutBuffer` should expose a `get_filled_batch(n)` API.
2. **Reconcile the docstring**: drop the 'wraps modulo' claim (the code's current monotonic-with-reset behavior is fine; just document it accurately).

## Reproduction

Write a unit test that creates a learner with `buffer_size = 64`, ingests only 10 experiences, and asserts the batch passed to PPO has length 10, not 64.

## Acceptance

- New regression test confirming partial-fill rollouts don't include zero-padded rows.
- `buffer_step_idx` docstring accurately describes the actual behavior.
- All existing tests still pass.

## Discovery

Flagged by the judge during PR #12 review ([approval comment](https://github.com/rjwalters/thrust/pull/12)) as non-blocking follow-ups.

rjwalters:
## Summary

- Restricts both PPO batch construction and GAE backward iteration to the actually-filled prefix of `RolloutBuffer` (`buffer_step_idx` rows), so zero-padded unwritten tail rows can no longer contaminate gradients.
- Adds `RolloutBatch::from_buffer_partial`, `RolloutBuffer::get_filled_batch`, and `compute_advantages_partial` as additive APIs. The existing `get_batch` / `compute_advantages` remain (their docstrings now flag the full-capacity semantics).
- Corrects the `buffer_step_idx` docstring on `PolicyLearner` — the old text claimed "wraps modulo `config.buffer_size`", which the code has never done; it is in fact monotonic with a single reset at the end of each successful `train()` cycle.

## Second contamination point (not in the original issue)

The original report only flagged `train_step`'s call to `get_batch()`. During curator verification, a second contamination point was identified and is also fixed in this PR: `PolicyLearner::train` calls `buffer.compute_advantages(...)` *before* `train_step`, and the old `compute_advantages` iterates the full `[num_steps, num_envs]` capacity with the inner backward GAE loop walking `(0..num_steps).rev()`. With `buffer_size = 64` and 10 filled rows, GAE iterates 64 steps backward starting from row 63 with `last_value = V(s_{T+1})` — but the bootstrap is correctly anchored after the 10th row, *not* after row 63. The 54 padding rows have `reward = 0`, `value = 0`, `terminated = false`, so `delta = γ * next_value`, and GAE accumulates `γ^k * last_value` backward through the padding before reaching the real data at index 9. Result: the 10 real rows get an advantage signal corrupted by 54 steps of decayed bootstrap value, and the 54 padding rows get nonzero advantages/returns that would also be sliced into the PPO batch.

The GAE fix is purely a loop-bound change, not a math change: the bootstrap value already corresponds to "the state after the last filled row", so clamping the inner loop to `(0..valid_steps).rev()` is correct. This is verified by `test_compute_advantages_partial_matches_full_on_prefix`, which asserts that partial-fill GAE over the first 3 rows of an 8-row buffer matches full GAE over a tight 3-row buffer to within 1e-6, and by `test_compute_advantages_partial_does_not_leak_bootstrap_through_padding`, which uses closed-form expected values.

Closes #28.

## Test plan

- [x] `cargo test --lib` — all 110 unit tests pass (8 new), including:
  - `test_train_step_does_not_zero_pad_partial_rollout` — end-to-end: 10 experiences with sentinel `old_log_prob = -0.69` into a `buffer_size = 64` learner. Asserts the PPO batch has exactly 10 rows and every `old_log_prob` in the batch equals the sentinel (would fail at `0.0` for any leaked zero-padded row).
  - `test_compute_advantages_partial_matches_full_on_prefix` — partial GAE on a prefix matches full GAE on a tight buffer.
  - `test_compute_advantages_partial_does_not_leak_bootstrap_through_padding` — direct regression for the second contamination point with closed-form expectations.
  - `test_rollout_batch_from_buffer_partial_skips_unfilled_tail` — batch construction correctness.
  - Plus `should_panic` tests for `valid_steps > num_steps` on both new APIs, and zero-valid-steps no-op tests.
- [x] All pre-existing tests still pass (`test_learner_ingests_experiences_and_trains`, etc.). Note: `tests/test_ppo_learning.rs::test_ppo_learns_from_synthetic_data` was already failing on master; not affected by this change.
- [x] `cargo +nightly clippy --lib --tests` — no new warnings on the changed files (`src/buffer/rollout.rs`, `src/buffer/rollout/gae.rs`, `src/buffer/rollout/storage.rs`, `src/multi_agent/learner.rs`).

### DIFF
--- a/src/buffer/rollout.rs
+++ b/src/buffer/rollout.rs
@@ -16,7 +16,8 @@
 
 // Re-export main components
 pub use gae::{
-    compute_advantages, compute_mc_returns, compute_nstep_returns, normalize_advantages,
+    compute_advantages, compute_advantages_partial, compute_mc_returns, compute_nstep_returns,
+    normalize_advantages,
 };
 pub use sampling::{
     Minibatch, MinibatchIterator, generate_minibatch_indices, sample_minibatch, shuffle_indices,
@@ -34,6 +35,9 @@ impl RolloutBuffer {
     /// Compute advantages using Generalized Advantage Estimation
     ///
     /// This is a convenience method that calls the module-level function.
+    /// Iterates the full `[num_steps, num_envs]` capacity; use
+    /// [`Self::compute_advantages_partial`] when the buffer is only
+    /// partially filled.
     ///
     /// # Arguments
     /// * `last_values` - Value estimates for the final states `[num_envs]`
@@ -43,10 +47,49 @@ impl RolloutBuffer {
         gae::compute_advantages(self, last_values, gamma, gae_lambda);
     }
 
+    /// Compute advantages over the first `valid_steps` rows only.
+    ///
+    /// Convenience wrapper around the module-level
+    /// [`compute_advantages_partial`]. Use this when the rollout buffer
+    /// has been partially filled (`valid_steps < num_steps`) to prevent
+    /// zero-padded tail rows from contaminating GAE on the real prefix.
+    ///
+    /// # Arguments
+    /// * `valid_steps` - Number of filled rows at the start of the buffer
+    /// * `last_values` - Bootstrap `V(s_{T+1})` for the state after row
+    ///   `valid_steps - 1`, per environment `[num_envs]`
+    /// * `gamma` - Discount factor
+    /// * `gae_lambda` - GAE lambda parameter
+    pub fn compute_advantages_partial(
+        &mut self,
+        valid_steps: usize,
+        last_values: &[f32],
+        gamma: f32,
+        gae_lambda: f32,
+    ) {
+        gae::compute_advantages_partial(self, valid_steps, last_values, gamma, gae_lambda);
+    }
+
     /// Get a batch of all data from the buffer
     ///
     /// This is a convenience method that calls the module-level function.
+    /// Returns the full `[num_steps, num_envs]` capacity, including any
+    /// zero-padded unwritten tail; use [`Self::get_filled_batch`] when
+    /// the buffer is only partially filled.
     pub fn get_batch(&self) -> RolloutBatch {
         RolloutBatch::from_buffer(self)
+    }
+
+    /// Get a batch covering only the first `valid_steps` rows.
+    ///
+    /// Use this when the buffer was filled with fewer than `num_steps`
+    /// transitions (e.g. an early-terminating rollout). It prevents the
+    /// zero-initialized tail of the buffer from being handed to PPO as
+    /// fake training data.
+    ///
+    /// # Panics
+    /// Panics if `valid_steps > self.shape().0`.
+    pub fn get_filled_batch(&self, valid_steps: usize) -> RolloutBatch {
+        RolloutBatch::from_buffer_partial(self, valid_steps)
     }
 }
--- a/src/buffer/rollout/gae.rs
+++ b/src/buffer/rollout/gae.rs
@@ -34,9 +34,53 @@ pub fn compute_advantages(
     gamma: f32,
     gae_lambda: f32,
 ) {
+    let num_steps = buffer.shape().0;
+    compute_advantages_partial(buffer, num_steps, last_values, gamma, gae_lambda);
+}
+
+/// Compute Generalized Advantage Estimation (GAE) over the first
+/// `valid_steps` rows of the buffer.
+///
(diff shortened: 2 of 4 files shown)


rjwalters:
## Judge review: Approve

Thorough verification of all claims; the fix is correct, well-tested, and minimal.

### Implementation verified
- **`RolloutBatch::from_buffer_partial`** (`src/buffer/rollout/storage.rs`) iterates only `valid_steps` rows; `from_buffer` correctly delegates with `valid_steps = buffer.num_steps`; panics with a descriptive message on overflow.
- **`compute_advantages_partial`** (`src/buffer/rollout/gae.rs`) restricts the per-env slices to `valid_steps` via `.take(valid_steps)` before invoking `compute_gae_single_env`. Since that helper derives its own `num_steps = rewards.len()`, the bootstrap branch (`t == num_steps - 1`) naturally fires at row `valid_steps - 1`. Zero-valid no-op and overflow assertion are present.
- **`RolloutBuffer::compute_advantages_partial` / `get_filled_batch`** are clean wrappers with appropriate re-exports.
- **`PolicyLearner::train` / `train_step`** now route through the partial variants (lines 151 and 309). The `buffer_step_idx` docstring is updated and no longer claims "wraps modulo `config.buffer_size`".

### GAE math reasoning validated
This is the critical claim, and it holds. `compute_bootstrap_last_values()` derives `last_value` from `self.last_experience.next_observation`, which `ingest_experience()` updates on every successful insert. After `valid_steps = 10` insertions, `next_observation` is the state that follows row 9 — exactly `V(s_{T+1})` with `T = valid_steps - 1`. The partial GAE walks backward from row 9 using that bootstrap, semantically identical to the full-buffer case. This is purely a loop-bound change, not a math change.

The `test_compute_advantages_partial_does_not_leak_bootstrap_through_padding` test pins this directly with a closed-form check: with two real rows and a bootstrap of 1.0, advantage[0] must equal `γ²λ * bootstrap`, not `γ^15 * bootstrap` (the contamination signature on a 16-row buffer). The assertion message itself diagnoses the failure mode — nice touch.

### Local validation
- `cargo test --lib`: 110 passed, 0 failed (matches builder's report); all four new GAE tests and three new batch tests pass.
- `cargo +nightly fmt --check`: clean.
- `cargo +nightly clippy --lib --tests`: zero new warnings in `src/buffer/rollout/*` or `src/multi_agent/learner.rs`.

### CI status
- `Check (no libtorch)` (both platforms), `Rustfmt`, `Documentation`, `Security Audit`: all pass.
- `Tests (Linux + libtorch)` and `Tests (macOS + libtorch)`: lib tests report 110/110 pass on both platforms. The only failure is `tests/test_ppo_learning.rs::test_ppo_learns_from_synthetic_data`, which is the pre-existing master flake explicitly flagged as out-of-scope for this review.

### Code quality
- Docstrings are precise and call out the contamination semantics in plain language.
- Tests cover the four orthogonal axes: math equivalence vs. tight prefix, the direct contamination case, the zero-step edge case, and the overflow guard.
- Backward compatibility is preserved — old APIs delegate to the new partial variants.
- No security or memory-safety concerns.

Ship it.

Issue #39, opened 2026-05-15
Audit unsafe pointer aliasing in PolicyLearner::train_step


rjwalters:
## Background

`PolicyLearner::train_step` in `src/multi_agent/learner.rs` contains an `unsafe` block (around lines 321-337 at time of filing, but verify — the file shifted in PRs #4 and #34) that uses raw pointer aliasing. Flagged as a non-blocking pre-existing concern by *two separate judges* during the recent batch:

- PR #12 (issue #4) approval comment: "Pre-existing `unsafe` pointer aliasing in `train_step()` (lines 321-337) untouched by this PR."
- PR #34 (issue #28) approval comment: noted again in the post-merge review.

Both reviews explicitly scoped it out ("untouched by this PR") but flagged it as worth a separate audit. It's never been verified since the recent learner.rs refactors (PR #12 fixed buffer-drop, PR #34 added partial-batch APIs).

## Investigation needed

1. **Find the `unsafe` block** — `grep -n unsafe src/multi_agent/learner.rs`. Document its current line range, what it does, and what invariants it relies on.
2. **Is it actually necessary?** `unsafe` blocks in Rust are usually needed for one of: FFI, raw allocation, manual lifetime extension, or escaping aliasing rules around `&mut`. Often a refactor lets you remove them.
3. **Are the safety invariants documented?** Rust convention: every `unsafe` block should have a `// SAFETY:` comment explaining why the unsafe operation is valid. If missing, that's a defensive-coding gap regardless of correctness.
4. **Does it survive PRs #12 and #34's changes?** The buffer plumbing and partial-fill paths added new code around this. Verify the aliasing assumptions still hold.

## Acceptance

One of:
- (Best) Remove the `unsafe` block via a refactor (split borrows, use `Cell`/`RefCell`, restructure indexing, etc.). All tests still pass.
- (Next best) Keep the `unsafe` block but add a thorough `// SAFETY:` comment justifying it. Reviewer agrees the invariants are sound.
- (Acceptable) Confirm the block is correct and well-justified as-is, and just add the `// SAFETY:` comment.

## See also

- PR #12 approval: https://github.com/rjwalters/thrust/pull/12
- PR #34 approval: https://github.com/rjwalters/thrust/pull/34
- #4 — original PolicyLearner cleanup
- #28 — partial-rollout fix

rjwalters:
## Curator Investigation

### 1. Current location of the `unsafe` block

[...]

PR #43, opened 2026-05-15
Fix nested PPO epoch loop in PolicyLearner::train_step

### LINKED ISSUE #41: PolicyLearner::train_step runs n_epochs^2 PPO epochs per call (nested epoch loop)
## Symptom

`PolicyLearner::train_step` in `src/multi_agent/learner.rs` runs an outer `for _ in 0..n_epochs { ... }` loop that calls into `PPOTrainer::train_step_with_aux` (or `train_step_with_policy`), which internally runs ANOTHER `for epoch in 0..n_epochs { ... }` loop over the same data. Net effect: every `train_step` call does `n_epochs²` PPO epochs instead of `n_epochs`.

With the codebase's `LearnerConfig::n_epochs = 4` (overriding the PPO default of 10 via `From<LearnerConfig>`), each call to `PolicyLearner::train_step` does **16 epochs**, not 4.

## Reproduction

Add an `eprintln!` at the top of `PPOTrainer::train_step_with_aux`'s inner epoch loop, run any test that exercises `PolicyLearner::train_step`, and observe the iteration count multiplied.

Or, by reading: open `src/multi_agent/learner.rs:302-370` and `src/train/ppo/trainer.rs:184-186`.

## Root cause

One of two outer/inner loops is redundant. Either:
- The outer loop in `PolicyLearner::train_step` was added before the inner loop in `PPOTrainer::train_step_with_aux` existed, and was never removed.
- Or the inner loop was added without noticing the outer one exists.

Either way, the current behavior is almost certainly not what either author intended.

## Why this matters

- **Throughput**: training takes ~4-16x longer per nominal step than declared.
- **Hyperparameter intent**: configured `n_epochs` doesn't mean what the user/scripts think it means.
- **Convergence behavior**: `n_epochs` per minibatch directly affects PPO's clipping behavior and gradient magnitudes — silently squaring it changes the algorithm's effective hyperparameters.

## Recommended fix

Drop the outer loop in `PolicyLearner::train_step`. The inner PPO loop is the load-bearing one — it's how PPO is supposed to work (multiple epochs over the same rollout). Update the surrounding logging/stats accumulation to reflect a single trainer call rather than `n_epochs` calls.

## Acceptance

- `PolicyLearner::train_step` runs exactly `config.n_epochs` PPO epochs (verified by counting via an instrumented build).
- Existing tests pass (the algorithmic change is correctness-positive — fewer redundant updates).
- Any benchmark numbers that were calibrated against the buggy behavior will shift; that's expected and welcome.

## Discovery

Flagged by the curator during investigation of #39 (`unsafe` audit) — out of scope for that issue, filed as its own. See [#39 curator comment](https://github.com/rjwalters/thrust/issues/39#issuecomment-4456739019).

rjwalters:
Closes #41

## Summary

`PolicyLearner::train_step` wrapped a call to `PPOTrainer::train_step_with_policy` in an outer `for _ in 0..n_epochs { ... }` loop. The trainer call itself already iterates over `config.n_epochs` PPO epochs internally, so every learner step was performing **n_epochs^2** PPO epochs over the same rollout.

With the default `LearnerConfig::n_epochs = 4` (which overrides PPO's default 10 via `From<LearnerConfig>`), this translated to 16 epochs per nominal step instead of 4 --- roughly 4x more gradient updates than declared, with corresponding effects on clipping behavior and gradient magnitudes.

## Curator's finding (from #39)

> `PolicyLearner::train_step` in `src/multi_agent/learner.rs` runs `for _ in 0..n_epochs { ... }` around a call into `PPOTrainer::train_step_with_aux` (or `train_step_with_policy`), which internally runs ANOTHER `for epoch in 0..n_epochs { ... }` over the same data. Net effect: `n_epochs^2` PPO epochs per call. With `LearnerConfig::n_epochs = 4` (overrides PPO default 10 via `From<LearnerConfig>`), this is 16 epochs per nominal step.
>
> The inner loop in `PPOTrainer::train_step_with_aux` is the **correct** PPO behavior --- multiple epochs over the same rollout is how PPO is supposed to work. The outer loop in `PolicyLearner::train_step` is the redundant one.

Reference: https://github.com/rjwalters/thrust/issues/39#issuecomment-4456739019

## Changes

- Removed the outer `for _ in 0..self.config.n_epochs` loop in `PolicyLearner::train_step`.
- Dropped the per-epoch stats accumulation + division-by-n_epochs averaging, since the trainer already returns stats averaged across all minibatch updates.
- Updated the docstring on `train_step` to clarify that it delegates a single trainer call which internally iterates `config.n_epochs` epochs.
- Preserved the existing `unsafe` block around the trainer call as-is. That is being addressed independently in #39 (the heads-up in the issue noted #39's builder is also touching this same function; merge ordering will need to be sorted manually).

## Test plan

- [x] `cargo build --features training` passes (locally with `LIBTORCH_USE_PYTORCH=1 LIBTORCH_BYPASS_VERSION_CHECK=1`).
- [x] `cargo test --features training --lib` --- all 109 tests pass.
- [x] No tests required adjustment: both `test_learner_ingests_experiences_and_trains` and `test_train_step_does_not_zero_pad_partial_rollout` use `n_epochs: 1` and only assert finiteness of returned stats, so neither was implicitly depending on the buggy behavior.

## Note on benchmarks

Any benchmark numbers calibrated against the old behavior may shift after this lands. That is expected --- the algorithm now does what was advertised (`config.n_epochs` PPO epochs per call, not `n_epochs^2`). Per-step throughput should improve by roughly `n_epochs`x (4x at the default), and convergence behavior may change because PPO's effective hyperparameters now match the configured ones.

### DIFF
--- a/src/multi_agent/learner.rs
+++ b/src/multi_agent/learner.rs
@@ -299,7 +299,13 @@ impl PolicyLearner {
         }
     }
 
-    /// Run PPO training step
+    /// Run a PPO training step.
+    ///
+    /// Delegates to `PPOTrainer::train_step_with_policy`, which internally
+    /// iterates over the rollout for `config.n_epochs` PPO epochs and returns
+    /// averaged statistics across all minibatch updates. This method performs
+    /// exactly one such trainer call --- it does NOT add an outer epoch loop
+    /// (doing so would compound epochs multiplicatively; see issue #41).
     fn train_step(&mut self) -> Result<TrainingStats> {
         // Get batch from the filled prefix of the buffer. `RolloutBuffer`
         // pre-allocates its full capacity at construction, so calling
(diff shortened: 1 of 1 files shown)


rjwalters:
## Approved

Surgical fix for the n_epochs² PPO bug. Verified against the curator's finding from #39 and the acceptance criteria.

### Code review

1. **Outer loop removed**: the `for _ in 0..self.config.n_epochs { ... }` wrapper around the trainer call is gone (was at lines 327-369). The trainer's internal `for epoch in 0..self.config.n_epochs` in `PPOTrainer::train_step_with_aux` at `src/train/ppo/trainer.rs:185` remains as the canonical PPO epoch loop, which is correct.

2. **Single trainer call preserved**: the call to `train_step_with_policy` with all seven tensor args plus the `evaluate_actions` closure is identical to the old loop body — no functional change inside the call.

3. **Stats math is correct**: confirmed by reading `src/train/ppo/trainer.rs:172-283`. The trainer accumulates `stats_sum` over every minibatch update (`stats_sum.add(&step_stats); num_updates += 1`) and returns `stats_sum.average()` — i.e., the trainer-returned stats are already averaged across all `n_epochs * n_minibatches` updates. The old code's outer-loop accumulator + divide-by-`n_epochs` was computing an average-of-averages, which the new code correctly replaces with a direct pass-through. The new path reports stats over the same number of underlying gradient updates that actually happened, so it is neither under- nor over-reporting.

4. **Docstring updated**: the new doc comment on `train_step` (lines 302-308) explicitly names the trap (`...doing so would compound epochs multiplicatively; see issue #41`) and explains that the trainer delegate already runs `config.n_epochs` epochs internally. Good defense against regression.

5. **Unsafe block preserved**: the raw pointer trainer/policy access (`trainer_ptr` / `policy_ptr` block) is retained unchanged, leaving #39's safe-borrow refactor cleanly scoped. The safety comment is preserved verbatim.

6. **`kl_divergence` field**: the multi-agent `TrainingStats` has a `kl_divergence` field that this code does not populate from the trainer's `approx_kl`. However, the old code did not populate it either (it relied on `TrainingStats::default()` zero-initialization), so this is not a regression introduced by the PR. Out of scope.

### Test impact

- `test_learner_ingests_experiences_and_trains` (line 519) — uses `n_epochs: 1`, immune.
- `test_train_step_does_not_zero_pad_partial_rollout` (line 623) — uses `n_epochs: 1`, immune.
- `tests/test_joint_trainer.rs`, `tests/test_ppo_learning.rs`, `tests/integration_train_export_inference.rs` — call `PPOTrainer` directly, not via `PolicyLearner::train_step`. Unaffected.
- No test was found in the repo that relied on the 16-epochs-per-call behavior. Nothing to flag for silent calibration against the bug.

### CI verification

All 7 jobs green:
- Rustfmt ✅
- Check (no libtorch) ubuntu-latest ✅
- Check (no libtorch) macos-latest ✅
- Tests (Linux + libtorch) ✅
- Tests (macOS + libtorch) ✅
- Documentation ✅
- Security Audit ✅

Merge state: `CLEAN`, `MERGEABLE`.

### Merge ordering note

This PR and #39 (the unsafe-block refactor) both modify `PolicyLearner::train_step` and will conflict at merge. Recommend merging #43 first — it is the smaller, more focused diff (47/-43 lines, all behavioral, no API changes) — and rebasing #39's safe-borrow refactor on top. This is a human merge-ordering call, not a blocker.

Approving.

Issue #58, opened 2026-06-05
DQN: add Double-DQN and soft (Polyak) target updates to reliably solve CartPole


rjwalters:
## Background

Vanilla DQN (landed in PR #52, closes #48) does not currently reach the CartPole ≥475 average reward threshold listed in #48's acceptance criteria. The smoke run climbed from ~20 to a peak of ~175 around 28k env steps, then exhibited the well-known catastrophic-forgetting dip back to ~120 by 60k steps. PR #52 was approved with this as an explicit deferral, on the grounds that the scaffolding is sound and reliably solving CartPole is downstream of the standard fixes.

## What this issue does

Land the two highest-leverage standard fixes to make CartPole reliable:

### 1. Double-DQN target computation
Change the TD target in `src/train/dqn/loss.rs` from

    y = r + γ · max_a' Q_target(s', a') · (1 - done)

to the Double-Q form:

    a* = argmax_a' Q_online(s', a')
    y = r + γ · Q_target(s', a*) · (1 - done)

This decouples action selection from action evaluation, killing the well-known overestimation bias that makes vanilla DQN unstable on CartPole.

### 2. Soft (Polyak) target updates
Add a `soft_update_tau: Option<f64>` field to `DQNConfig`. When `Some(tau)`, replace the hard target-net copy in `maybe_sync_target` with `target_params = tau * online_params + (1 - tau) * target_params`, applied every step (not every `target_update_interval` env steps). Default to `None` for backwards compatibility; the CartPole example flips it on.

## Acceptance criteria

- `DQNConfig` gains `soft_update_tau: Option<f64>` (default `None`).
- `compute_td_target` in `src/train/dqn/loss.rs` uses Double-Q indexing.
- `DQNTrainer::maybe_sync_target` does Polyak when `soft_update_tau.is_some()`, else falls back to today's hard copy.
- Unit tests: Double-Q target matches a hand-computed value on a 2-step synthetic; soft update with `tau=0.005` moves target params toward online params (not equal to them).
- `examples/games/cartpole/train_cartpole_dqn.rs` configures `soft_update_tau(0.005)` and reaches **≥475 average reward over the last 100 episodes** within ~50k env steps on a fixed seed. (The actual deferred acceptance criterion from #48.)
- README "Algorithms" section mentions Double-DQN.

## Out of scope (file separate issues if needed)

- Prioritized Experience Replay (see follow-up issue)
- Dueling Q-network architecture
- n-step returns
[...]

PR #66, opened 2026-06-05
fix(dqn): remove misleading target_synced field from DQNStepStats

### LINKED ISSUE #57: DQN: DQNStepStats::target_synced is always false (cosmetic bug)
## Bug

In `src/train/dqn/trainer.rs`, the `DQNStepStats` struct returned by `DQNTrainer::train_step` hard-codes `target_synced: false` with a comment like `// set by caller via maybe_sync_target`, but no caller ever patches it. The example loop calls `maybe_sync_target()` correctly on the env-step counter, so the algorithm is correct — only the diagnostic field is misleading. Tooling that reads `stats.target_synced` will always see `false`.

## Fix

One of:

1. Move the sync-counter logic INTO `train_step` and have it set the field correctly. (Changes the call-shape of the trainer; nontrivial.)
2. Remove the `target_synced` field from `DQNStepStats` entirely and let callers track sync timing themselves. (Simpler.)
3. Plumb the sync result back from `maybe_sync_target` so the caller can update the field before logging. (Awkward.)

Option 2 is probably the cleanest — the field has no consumer today and adds nothing once removed.

## Acceptance

- `DQNStepStats::target_synced` either gets set correctly or is deleted.
- `cargo test --features training` passes.
- Example `examples/games/cartpole/train_cartpole_dqn.rs` still produces sensible output.

## Source

Surfaced by Judge during PR #52 review. Non-blocking at the time; logged as a "small follow-up tidy."

rjwalters:
## Summary

Removes the `target_synced: bool` field from `DQNStepStats` in `src/train/dqn/trainer.rs`. The field was always hardcoded to `false` with a comment claiming the caller would patch it, but no caller ever did. The env-step loop already tracks sync timing itself via `total_env_steps % target_update_interval`, so the field added nothing and was misleading to any tooling reading it.

This is option 2 from the issue (cleanest), as the field had no consumer.

## Scope of changes

- `src/train/dqn/trainer.rs`: removed the field from the `DQNStepStats` struct definition (with its doc comment) and from the struct literal returned by `train_step`. Total: 3 lines deleted, no other files touched.

No tests, examples, or other consumers referenced `target_synced` (verified by `grep -rn "target_synced"`), so no caller-side cleanup was needed.

## Test plan

- [x] `cargo build --features training --release` succeeds
- [x] `cargo test --features training` passes (153 lib tests + integration + doc tests, all green)
- [x] `cargo +nightly fmt --check` clean
- [x] `cargo +nightly doc --no-deps --all-features` produces 0 warnings
- [x] `examples/games/cartpole/train_cartpole_dqn.rs` builds

Closes #57

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/train/dqn/trainer.rs
+++ b/src/train/dqn/trainer.rs
@@ -27,8 +27,6 @@ pub struct DQNStepStats {
     pub epsilon: f64,
     /// Replay buffer fill level at the time of this update.
     pub buffer_len: usize,
-    /// Whether the target network was synced this step.
-    pub target_synced: bool,
 }
 
 /// DQN Trainer.
@@ -247,7 +245,6 @@ impl DQNTrainer {
             mean_q: mean_q_val,
             epsilon: self.last_epsilon,
             buffer_len: self.buffer.len(),
-            target_synced: false, // set by caller via maybe_sync_target
         }))
     }
 }


rjwalters:
## Judge Review: Approved

**Decision**: Approve — clean cosmetic cleanup, exactly as scoped by issue #57 option 2.

### Verification

- **Diff**: 3 lines deleted from `src/train/dqn/trainer.rs` only (field decl + doc comment + struct-literal assignment). No other files touched. Matches PR description.
- **Base**: `main`, mergeStateStatus `CLEAN`.
- **CI**: All 7 checks green (Tests Linux+libtorch, Tests macOS+libtorch, Check no-libtorch x2, Documentation, Rustfmt, Security Audit).
- **No remaining consumers**: builder's `grep -rn target_synced` confirms zero references in `src/`, `examples/`, or `tests/` post-PR.
- **Rubric fit**: issue #57 explicitly preferred option 2 ("field has no consumer and adds nothing once removed") — that's exactly what landed.

### Coordination with PR #67

PR #67 (Double-DQN + soft target updates, closes #58) touches the same area but the two are non-conflicting:

- PR #67's `dqn.rs` change to `DQNStepStats` is only a re-export line (`pub use trainer::{DQNStepStats, DQNTrainer}`) — no fields edited.
- PR #67 does not touch the `DQNStepStats` struct body or the `train_step` return literal where `target_synced` lived.
- `grep target_synced` on PR #67's diff: zero hits.

Whichever PR merges first, the other should rebase cleanly. The builders coordinated correctly.

### Notes for the breaking-change ledger

This is a public-API field removal on `DQNStepStats`. Pre-1.0, and the field was documented as caller-patched but never actually patched, so removing it is the right call — but downstream tooling that destructured `DQNStepStats { target_synced, .. }` will need a one-line edit. Worth a CHANGELOG line whenever the next release goes out.

Approving and handing off to merge.

PR #205, opened 2026-06-21
PSRO: per-iteration on_iteration callback + live tracing (NFSP-parity)

rjwalters:
Closes #202. PR B of the PSRO perf epic #198.

## What

Adopt NFSP's observability pattern for PSRO so per-iteration progress is visible **during** `run()` rather than only after the aggregate stats return.

### Callback API

```rust
pub fn run<F>(&mut self, mut on_iteration: F) -> Result<PsroStats>
where
    F: FnMut(&PsroIterationStats);

pub fn run_silent(&mut self) -> Result<PsroStats> { self.run(|_| {}) }
```

`on_iteration(&iter_stats)` fires once per outer iteration, immediately after that iteration's `PsroIterationStats` is constructed and before it is pushed to history. The `iteration` field increases monotonically `1..=max_iterations`. This mirrors `NfspTrainer::run` / `run_silent` exactly.

### Changes

- `src/multi_agent/psro.rs`: `run()` -> generic `run<F>(on_iteration)`; new `run_silent()` shim. In-module tests migrated to `run_silent()`.
- `examples/games/bucket_brigade/train_psro.rs`: live `iter k/N  pop_size=...  exploitability=...` logging via the callback; the "PSRO's run() is opaque" post-hoc stats-walk workaround removed. (Mid-run checkpoint **writing** remains #204; this PR only fires the callback per-iteration and wires live logging.)
- Integration tests migrated to `run_silent()`: `test_psro_matching_pennies`, `test_psro_n_player_matching_pennies`, `test_seeded_reproducibility`, `test_psro_bucket_brigade`.

### Tests

New unit tests in `psro.rs`:
- `test_psro_run_callback_fires_per_iteration` — asserts the callback fires exactly `max_iterations` times with monotonically increasing iteration indices matching the pushed history order.
- `test_psro_run_silent_records_full_history` — asserts `run_silent()` records the full per-iteration history.

## Backward compatibility

All existing `run()` callers/tests migrated to `run_silent()` in the same PR. Existing PSRO behaviour is unchanged (the no-op callback path is identical to the old loop).

## Coordination note

#201 (parallel PR) also edits `psro.rs` (payoff-eval seeding). My edits are localized to the `run()`/callback surface to minimize conflict.

## Verification

- `cargo test --features training --lib psro`: 23 passed, 0 failed (incl. 2 new tests).
- `cargo fmt --check`: clean.
[...]

Issue #204, opened 2026-06-21
PSRO: mid-run checkpointing during run() + parallelized-baseline calibration on alc-2 (operator-gated)


rjwalters:
**Part of #198. Depends on #203 and #202. OPERATOR-GATED — requires alc-2 (32-core) cluster hardware; a sweep Builder must NOT claim this (mirrors #134 and #184).**

## Background

The #134 cluster run found PSRO yields nothing on a killed run: `PsroTrainer::run()` had no mid-run checkpointing, so a multi-hour run that died left no artifact (`docs/research/2026-06-bucket-brigade-validation.md`). With the per-iteration `on_iteration` callback from #202 and the parallelized payoff eval from #203 in place, we can both checkpoint *during* the run and produce a real calibration.

### 1. Mid-run checkpointing (CI-testable part)

Using the `on_iteration` callback (#202), have `examples/games/bucket_brigade/train_psro.rs` write per-agent BR policies + the meta-Nash distribution to disk **during** `run()` at `CHECKPOINT_INTERVAL_ITERATIONS`, not after. The trainer stays format-agnostic (file IO + the `Recorder` live in the example, behind the `training` feature); expose whatever minimal accessor the closure needs to pull the newest BR per agent (e.g. reuse/extend `PsroTrainer::populations(agent)`). A killed run must leave resumable `psro_{cell}_iter{k}_agent{a}.bin` files.

### 2. Calibration on alc-2 (operator-gated part)

On a 32-core alc node (release, CPU NdArray), run the parallelized PSRO on at least one bucket-brigade no-convergence cell and capture:
- per-iteration wall-clock at the parallelized baseline,
- core utilization (confirm >1 core used),
- speedup vs the single-thread baseline,
- a feasible #134 PSRO budget recommendation.

## Acceptance criteria

- [ ] Intermediate checkpoints written during `run()` (resumable / partial-result-safe); a unit/integration test exercises the checkpoint-callback logic on CPU CI.
- [ ] **[operator]** A multi-core PSRO run on alc-2 showing >1 core utilized and a measured speedup vs the single-thread baseline, numbers committed to `docs/research/2026-06-bucket-brigade-validation.md`.
- [ ] **[operator]** A documented per-iteration wall-clock + a feasible #134 PSRO budget recommendation appended to the research doc / #134.

## Why operator-gated (do not relabel loom:issue)

[...]

Issue #202, opened 2026-06-21
PSRO: per-iteration on_iteration callback + live tracing (NFSP-parity), remove opaque-run() workaround


rjwalters:
**Part of #198.** Independent of #201 / the parallelization work — can be built in parallel.

## Background

`PsroTrainer::run()` (`src/multi_agent/psro.rs:1280`) is opaque: it runs all iterations internally and only returns the full `PsroStats` history at the end. The bucket-brigade example documents the pain directly — "PSRO's `run()` is opaque — we cannot easily inject a per-iter" (`examples/games/bucket_brigade/train_psro.rs:253`) — and works around it by walking the stats history *after* `run()` returns. Result: no live progress during multi-hour runs (the #134 cluster run showed "zero iteration output"), per `docs/research/2026-06-bucket-brigade-validation.md`.

NFSP already solved this cleanly: `NfspTrainer::run<F>(&mut self, mut on_iteration: F)` with a `run_silent()` shim (`src/multi_agent/nfsp.rs:527` / `:592`), and `train_nfsp.rs` logs `iter k/N reservoir_sizes=... avg_ap_loss=...` live (`train_nfsp.rs:241`). PSRO should match this exactly.

## Task

- Change `PsroTrainer::run()` to `run<F: FnMut(&PsroIterationStats)>(&mut self, on_iteration: F)`, invoking `on_iteration(&iter_stats)` immediately after each iteration's `PsroIterationStats` is constructed (psro.rs:1350), before pushing to history.
- Add `run_silent(&mut self) -> Result<PsroStats>` delegating to `run(|_| {})` (mirrors NFSP).
- **Backward-compat:** migrate existing no-arg `run()` callers/tests to `run_silent()` in the same PR (matching-pennies tests, any others).
- Update `examples/games/bucket_brigade/train_psro.rs` to use the callback for live `tracing::info!` per-iteration logging matching NFSP's format (e.g. `iter k/N  exploitability=...  pop=...`), and **delete** the opaque-run post-hoc stats-walk workaround (train_psro.rs:253-312, keeping the final-save logic).

## Acceptance criteria

- [ ] `PsroTrainer::run` takes an `FnMut(&PsroIterationStats)` callback; `run_silent()` exists.
- [ ] A unit test asserts the callback is invoked exactly `max_iterations` times with monotonically increasing `iteration` values.
[...]

Issue #201, opened 2026-06-21
PSRO: make payoff-cell evaluation a pure, per-cell-seeded function (determinism prep)


rjwalters:
**Part of #198.**

## Background

`PsroTrainer::evaluate_payoff_joint` (`src/multi_agent/psro.rs:1591`) is the per-cell payoff evaluator that the `population^N` payoff-tensor loop (`run()`, psro.rs:1330-1340) calls for every boundary cell. It is the dominant cost in long PSRO runs (the #134 cluster run never completed 4 iters / 256 cells in 40 min — see `docs/research/2026-06-bucket-brigade-validation.md`).

Before that loop can be parallelized (#198 PR B), the evaluator must become **pure and order-independent**. Today it threads the **shared `&mut self.rng`** into `get_action_host_seeded` for action sampling (psro.rs:1622). That makes each cell's action draws depend on global RNG state and evaluation order, so the cells are NOT independent — a `par_iter` would silently change results.

The env-reset seed is *already* derived deterministically per `(joint, ep)` (psro.rs:1603-1610); only the action-sampling RNG is shared.

## Task

Refactor `evaluate_payoff_joint` so each cell evaluation is fully self-contained:

- Extract the body into a function that takes **no `&mut self.rng`**. It constructs a **local** `StdRng::seed_from_u64(per_cell_seed)` where `per_cell_seed` is derived from `(config.seed, joint_hash)` (reuse the existing little-endian `joint_hash` scheme), and threads that local RNG into both `reset_joint` and every `get_action_host_seeded` call.
- Per-episode seeding stays as-is (seed mixes in `ep`).
- The function must depend only on `(joint, config, the per-agent policies for that joint, env_factory)` — no other trainer mutable state.

This is a **behavior change** vs. the current shared-RNG path (the payoff numbers will move once), so it lands serially here, ahead of parallelization, with the new values pinned by tests.

## Acceptance criteria

- [ ] `evaluate_payoff_joint` no longer reads or mutates `self.rng`; action sampling uses a per-cell local `StdRng` seeded from `(config.seed, joint)`.
- [ ] A unit test asserts **order-independence**: evaluating the same set of boundary cells in two different orders (and re-evaluating a single cell twice) yields bit-identical payoff vectors.
[...]

PR #207, opened 2026-06-21
Parallelize PSRO payoff-tensor eval with rayon, bit-identical to serial

### LINKED ISSUE #203: PSRO: parallelize payoff-tensor evaluation with rayon, bit-identical to serial path
**Part of #198. Depends on #201.**

## Background

The PSRO payoff-tensor evaluation is the `population^N` bottleneck that made PSRO impractical on the #134 cluster run (4 iters / 256 cells never finished in 40 min; see `docs/research/2026-06-bucket-brigade-validation.md`). The boundary-slab loop in `PsroTrainer::run()` (`src/multi_agent/psro.rs:1330-1340`) evaluates every new joint-strategy cell **serially** on one core, ignoring the 31 other cores on the alc nodes.

After #201 makes `evaluate_payoff_joint` a **pure, per-cell-seeded** function (no shared `self.rng`), the boundary cells become independent and the loop is embarrassingly parallel. The repo already uses rayon under the `training` feature (`src/env/pool.rs:23` / `:98`, `Cargo.toml:72,127`) — follow that convention.

## Task

- Replace the serial boundary-slab loop (psro.rs:1332-1340) with a rayon parallel evaluation: collect the boundary `components` vectors, then `boundary.into_par_iter().map(|joint| (joint, evaluate_payoff_joint_pure(&joint, ...))).collect()`, then write results into the cache serially (deterministic insert order). Each task clones the joint's per-agent policies (`JointPolicy: Clone`) and builds its own env via `env_factory` — no shared mutable state crosses threads.
- Add the minimal `Send`/`Sync` bounds needed on the parallel path (mirror `EnvPool`'s `E: Send` convention, pool.rs:58/96). If trait-bound friction is high, gate parallelism behind the `rayon`/`training` feature with a serial fallback.
- **Determinism is the load-bearing guarantee:** because each cell seeds its own `StdRng` from `(config.seed, joint)` (#201), the parallel result MUST be bit-identical to the serial result regardless of thread count or scheduling.

## Acceptance criteria

- [ ] Boundary-cell payoff evaluation runs via rayon `par_iter` (under the `training`/`rayon` feature).
- [ ] A unit test asserts **par == serial bit-identically**: run the payoff tensor through the parallel path and a serial reference path and assert cell-for-cell equality (locally testable on CPU CI; does NOT require multi-core hardware to validate correctness — rayon will use available threads and the seeding makes the result thread-count-invariant).
- [ ] Existing PSRO determinism tests still pass.
- [ ] `cargo test --features training` green; no clippy regressions.

## Out of scope (noted in #198)

- Parallelizing the per-agent BR rollouts (smaller win: N vs population^N; larger determinism refactor since BR shares `self.populations` growth + opponent-sampling RNG). Stretch note only.
- Committed wall-clock speedup numbers — those are operator-gated in #198 PR D (needs alc-2 32-core box).

## Affected files

- `/Users/rwalters/GitHub/thrust/src/multi_agent/psro.rs`

tier:goal-advancing

rjwalters:
Closes #203. PR C of the PSRO perf epic #198 (depends on #201, merged).

## What

The PSRO boundary-slab loop in `PsroTrainer::run()` evaluated every new joint-strategy cell serially on one core — the `population^N` bottleneck from the #134 cluster run. Since #201 made `evaluate_payoff_joint_pure` a pure, per-cell-seeded function (local `StdRng` from `(config.seed, joint)`, no shared trainer RNG), the slab is now embarrassingly parallel.

## Changes

- **`evaluate_payoff_boundary_parallel`**: `par_iter`s over the boundary cells, clones each joint's per-agent policies (`P: Clone`), builds a fresh env via `env_factory`, and calls `evaluate_payoff_joint_pure`. Results are collected by index (rayon preserves input order), then written into the cache serially in the same deterministic order as the old loop.
- **`run()` loop**: gathers the boundary cells in deterministic flat order, then dispatches them to the parallel evaluator.

## Send/Sync handling

The parallel closures bind only the `Sync` field borrows (`&self.populations`, `&self.config`, `&self.env_factory`, `&self.device`) into locals, so they never capture the whole `&self` — which also holds the non-`Sync` `Box<dyn MetaSolver>` and the `FP`/`FO` factory closures. Minimal bounds `P: Send + Sync`, `E: Send`, `FE: Sync`, `B::Device: Sync` are added to `run`/`run_silent` only, mirroring the `EnvPool` Send-bound convention (`src/env/pool.rs:58`). **No `Mutex`** is introduced, so the hot loop is never serialized.

## Bit-identity (load-bearing)

New test `test_payoff_boundary_parallel_matches_serial_bit_identically`:
- asserts the parallel slab equals a serial cell-by-cell reference **cell-for-cell exactly**, and
- re-runs the parallel path inside rayon thread pools of size **1 and 4**, asserting both equal the serial reference — proving thread-count invariance. Fully CPU-CI-testable; no cluster needed.

## Pre-existing breakage fixed

`test_payoff_cell_eval_is_order_independent` still called the no-arg `run()` removed by #202 (#205) — origin/main did not compile its psro lib tests. Switched that call to `run_silent()`.

## Out of scope

Per-agent BR rollout parallelism (smaller win, larger determinism refactor) and committed wall-clock speedup numbers (operator-gated #204).

## Verification

- `cargo test --features training --lib psro`: 26 passed.
- `cargo test --release --features training --test test_psro_matching_pennies`: 2 passed (determinism/convergence unchanged).
- `cargo fmt --check`: clean.
- `cargo clippy --all-targets --features training -- -D warnings`: zero new errors in touched files (pre-existing repo-wide clippy debt untouched).
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --features training`: clean.

### DIFF
--- a/src/multi_agent/psro.rs
+++ b/src/multi_agent/psro.rs
@@ -92,6 +92,7 @@
 use anyhow::{Result, anyhow};
 use burn::{optim::Optimizer, tensor::backend::AutodiffBackend};
 use rand::{Rng, SeedableRng, rngs::StdRng};
+use rayon::prelude::*;
 
 use crate::{
     multi_agent::joint::{
@@ -1294,6 +1295,14 @@ where
     pub fn run<F>(&mut self, mut on_iteration: F) -> Result<PsroStats>
     where
         F: FnMut(&PsroIterationStats),
+        // Bounds required by the rayon-parallel boundary-slab evaluation
+        // (issue #203). Mirror the `EnvPool` Send-bound convention
+        // (pool.rs:58). The parallel payoff result is bit-identical to a
+        // serial sweep because each cell is pure (issue #201).
+        P: Send + Sync,
+        E: Send,
+        FE: Sync,
+        B::Device: Sync,
     {
         let num_agents = self.joint_config.num_agents;
 
@@ -1346,14 +1355,31 @@ where
             // new radix. We iterate flat indices and decompose.
             let total_new = new_k.checked_pow(num_agents as u32).expect("k^N overflow");
             let new_strategy_idx = new_k - 1;
-            for s in 0..total_new {
-                let components = decompose_joint_index(s, num_agents, new_k);
-                let touches_boundary = components.contains(&new_strategy_idx);
-                if !touches_boundary {
-                    continue;
-                }
-                let payoffs = self.evaluate_payoff_joint(&components);
-                self.payoff_cache.set_cell(&components, payoffs);
+
+            // Gather the boundary cells (those whose per-agent index
+            // vector includes the newest strategy) in deterministic flat
+            // order. This is the `population^N` slab that dominates PSRO
+            // cost on large N (issue #198).
+            let boundary: Vec<Vec<usize>> = (0..total_new)
+                .filter_map(|s| {
+                    let components = decompose_joint_index(s, num_agents, new_k);
+                    components.contains(&new_strategy_idx).then_some(components)
+                })
+                .collect();
+
+            // Evaluate every boundary cell in parallel. Each cell is a
+            // pure function of `(config.seed, joint)` (issue #201), so
+            // the parallel result is **bit-identical** to a serial sweep
+            // regardless of thread count or scheduling: cells share no
+            // mutable state, each clones its joint policies and builds a
+            // fresh env via `env_factory`, and seeds a local `StdRng`.
+            // Results are collected by index (not push order), then
+            // written into the cache serially below, so the cache is
+            // populated in the same deterministic order as the old serial
+            // loop. See `evaluate_payoff_boundary_parallel`.
+            let evaluated = self.evaluate_payoff_boundary_parallel(&boundary);
+            for (components, payoffs) in boundary.iter().zip(evaluated) {
+                self.payoff_cache.set_cell(components, payoffs);
             }
 
             // Step 4: re-solve the meta-Nash on the post-append cache
@@ -1381,7 +1407,13 @@ where
     /// iteration callback. Use this when per-iteration observation is
     /// not needed (mirrors
     /// [`NfspTrainer::run_silent`](crate::multi_agent::nfsp::NfspTrainer::run_silent)).
-    pub fn run_silent(&mut self) -> Result<PsroStats> {
+    pub fn run_silent(&mut self) -> Result<PsroStats>
+    where
+        P: Send + Sync,
+        E: Send,
+        FE: Sync,
+        B::Device: Sync,
+    {
         self.run(|_| {})
     }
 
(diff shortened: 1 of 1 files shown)


rjwalters:
## Judge verdict: APPROVED

Reviewed with extra scrutiny on parallelism correctness. Verified locally (full Tests CI ~1-2h is intentionally not blocked per epic policy).

### Order independence (correct)
`evaluate_payoff_boundary_parallel` collects via `par_iter().map(...).collect()` — rayon preserves input order — and the caller writes the cache with `boundary.iter().zip(evaluated)` in the same deterministic flat order as the old serial loop. No push-order, no index races. The payoff tensor is order-independent.

### Bit-identity (the core bar) — holds
Each cell delegates to the pure `evaluate_payoff_joint_pure` (#201/#206), which seeds a local `StdRng` from `config.seed ^ splitmix64(joint_hash)` and per-episode reset seeds from `(config.seed, joint_hash, ep)` — zero shared mutable RNG. The new test `test_payoff_boundary_parallel_matches_serial_bit_identically` asserts parallel == serial cell-for-cell with `assert_eq!` on `Vec<Vec<f32>>` (exact, not approximate), and re-runs the parallel path inside dedicated rayon pools of size **1 and 4**, asserting both equal the serial reference — genuinely proving thread-count invariance.

### Send/Sync (minimal + correct)
The new fn binds only the Sync field borrows (`&populations`, `&config`, `&env_factory`, `&device`) into locals before `par_iter`, so the closure never captures the whole `&self` (which holds the non-Sync `Box<dyn MetaSolver>` and the `FP`/`FO` factory closures). Bounds `P: Send + Sync, E: Send, FE: Sync, B::Device: Sync` added only to `run`/`run_silent`/the new fn — mirroring the `EnvPool` convention. Each task clones its policies (`P: Clone`) and builds a fresh env. **No `Mutex`** — the hot loop is never serialized.

### Pre-existing broken-test fix (verified)
Confirmed origin/main's psro lib test target did NOT compile: `test_payoff_cell_eval_is_order_independent` called the no-arg `run()`, but `run` on main takes a `FnMut` callback (replaced in #205). #207 correctly switches it to `run_silent()`. **Merging this restores green psro tests.**

### Local verification (all green)
- `cargo test --features training --lib psro` — **26 passed**, 0 failed (compiles + passes, target restored)
- `cargo test --release --test test_psro_matching_pennies` — 2 passed
- `cargo test --release --test test_psro_n_player_matching_pennies` — 2 passed (the N-player `population^N` target)
- `cargo test --release --test test_seeded_reproducibility` — 2 passed (bit-identical traces)
- `cargo fmt --check` — clean
- clippy — zero warnings referencing psro.rs (the 45 errors are pre-existing debt in `src/inference/mod.rs`, untouched since #99; not introduced here)
- `RUSTDOCFLAGS=-D warnings cargo doc --no-deps` — clean

### Scope
Additive/contained: only `src/multi_agent/psro.rs` (+225/-10).

Approving on local-pass. UNSTABLE merge state is solely the still-pending long Tests CI.

Issue #198, opened 2026-06-20
PSRO perf + observability pass (blocks long-budget runs for #134)


rjwalters:
## Background

While executing the #134 research validation (long-budget PSRO/NFSP on the bucket-brigade no-convergence cells) on the alcubierro cluster (i9-14900K + RTX 4090 nodes), the **PSRO path proved impractical for any meaningful budget** with the current implementation. NFSP ran fine (~145s/iter at 2048 rollout, linear, observable); PSRO did not.

## Measured evidence (alc-2, release build, CPU NdArray)

| Probe (PSRO, beta01) | Result |
|----------------------|--------|
| 6 iters × 2048 | **>85 min, never completed**, zero iteration output |
| 4 iters × 2048 | **>40 min, never completed**, zero iteration output |

For comparison, NFSP on the same host/cell did 12 iters × 2048 end-to-end in ~29 min with clean per-iteration logging.

## Root causes (all in the PSRO path)

1. **Single-threaded.** `src/multi_agent/psro.rs` has no rayon/parallelism; runs pin one core at 99% and ignore the other 31. The 32-core nodes give zero speedup.
2. **α-rank payoff tensor scales as `population^num_agents` (= `population^4` here).** Each entry needs a rollout evaluation (`payoff_eval_episodes`), so cost is dominated by the last iterations and grows superlinearly. But even 4 iterations (4^4=256 evals) never completed, so the **per-iteration base cost is also very high** (rollout-based payoff eval on a slow env + BR training).
3. **`PsroTrainer::run()` is opaque.** The `train_psro` example notes it "cannot easily inject a per-iter" hook — it runs ALL iterations internally and only logs per-iteration stats + writes checkpoints *after* `run()` returns. So there is **no progress observability and no mid-run checkpointing**: a killed run yields nothing, and you can't tell which iteration it's on.

## Why this blocks #134 (PSRO half)

The #134 issue assumed a "multi-hour" run by linearly extrapolating the integration test's claimed "12 iters × 2048 in ~5 min" (~25s/iter). That rate does **not reproduce** — the real per-iteration cost is ~30×+ higher, and right-sizing the budget can't fix a structural single-threaded + opaque + superlinear-cost problem. NFSP is being run to deliver the tractable half of #134; PSRO is parked behind this issue.

## Proposed work

[...]

Issue #134, opened 2026-06-17
Bucket-brigade research validation: long-budget PSRO/NFSP training + evaluation on alc-2 (operator action required)


rjwalters:
> ## ⚠️ Update (2026-06-22): apply the #199/#215 convergence knobs in the re-run
>
> Since this issue was filed, the NFSP/PSRO non-convergence was root-caused and **code-level fixes landed** (the cluster-validation arms of #199 and #215, both now closed). The re-run below **must set the new knobs**, or it will reproduce the original negative result:
>
> - **NFSP (#199, PR #223)** — set `AP_COVERAGE` (adaptive average-policy supervised-step floor; default off) and `BR_REWARD_SCALE` (default 1.0). The `[-700,0]` payoff band needs scaling: `br_reward_scale≈0.01` cut BR critic `value_loss` ~10,000× locally. Writeup: `docs/research/2026-06-nfsp-avg-policy-undertraining.md`.
> - **PSRO (#215, PR #225)** — set `ALPHA_RANK_NORMALIZE_SPAN=1` (the α-rank Moran process saturates at this magnitude: α·δ≈7000 vs ≤20 on validated matching-pennies) and `BR_REWARD_SCALE`. Local 3-iter beta01 went from non-monotone → monotonically decreasing exploitability with both on. Writeup: `docs/research/2026-06-psro-alpha-rank-payoff-magnitude.md`.
>
> Example invocation (per cell, both trainers), extending the commands below:
> ```bash
> CELL=beta05 TOTAL_ITERATIONS=48 ROLLOUT_STEPS=8192 \
>   AP_COVERAGE=1.0 BR_REWARD_SCALE=0.01 \
>   cargo run --release --example train_nfsp --features "training,env-bucket-brigade" 2>&1 | tee nfsp_beta05.log
>
> CELL=beta05 TOTAL_ITERATIONS=48 ROLLOUT_STEPS=8192 \
>   ALPHA_RANK_NORMALIZE_SPAN=1 BR_REWARD_SCALE=0.01 \
>   cargo run --release --example train_psro --features "training,env-bucket-brigade" 2>&1 | tee psro_beta05.log
> ```
> (Confirm exact env-var names/defaults against the current `train_nfsp.rs` / `train_psro.rs` — these knobs were added after this issue's original body. All default to no-op, so they MUST be set explicitly.)
>
[...]

PR #214, opened 2026-06-21
PSRO: deterministic boundary-slab subsampling cap (closes #212)

### LINKED ISSUE #212: PSRO: cache/subsample the population^N payoff tensor (parallelism alone insufficient for long runs)
## Finding (from the #198 cluster calibration)

After #198 parallelized the PSRO payoff-tensor evaluation (rayon, bit-identical, #207), a calibration on alc-2 (32-core i9-14900K, 33 threads active, confirmed via `nproc`/thread count) shows the parallelization **works but is not sufficient** for long PSRO runs on the 4-player bucket-brigade game:

| iter | population | α-rank payoff evals (pop^4) | wall-clock |
|------|-----------|------------------------------|-----------|
| 1 | 2 | 16 | ~180 s |
| 2 | 3 | 81 | ~472 s |
| 3 | 4 | 256 | >620 s (killed mid-iter) |

Per-iteration cost grows ~2.6×/iteration. The payoff tensor is `population^num_agents = population^4`, so iter 12 (pop 13) would need **28,561 evaluations** — a constant ~33× core speedup cannot beat super-linear growth. A 12-iteration run projects to many hours-to-days; the #134-scale budget is infeasible even parallelized.

(For contrast: pre-#198 single-threaded PSRO couldn't complete 4 iterations in 40 min and emitted no progress. #198 delivered a real, observable, bit-identical speedup on early iterations — it just doesn't change the asymptotics.)

## The actual lever: cache reusable payoff entries

PSRO adds exactly **one new policy per agent per outer iteration**. The vast majority of the `population^4` payoff-tensor entries are between policies that already existed last iteration and are therefore **unchanged** — only the entries involving at least one brand-new policy need (re)evaluation. Caching the payoff tensor across iterations turns each iteration's NEW work from `O(pop^4)` into `O(pop^3)` (the slabs touching the new policy), a large asymptotic win that compounds with the existing parallelism.

Suggested work:
1. Persist the payoff tensor across outer iterations; on each iteration, evaluate only the entries involving a newly-added policy (the "boundary slabs"), reusing cached values for the rest. (The code already evaluates a "boundary" per iteration — verify whether it's recomputing interior cells.)
2. Optionally add a `max_population_size` / payoff-subsampling cap to bound worst-case cost.
3. Re-run the alc-2 calibration after caching; recompute a feasible #134 PSRO budget.

## Secondary observation (separate concern)
Exploitability *increased* across the calibration iterations (6404 → 8864), i.e. PSRO is not converging on bucket-brigade — sibling to the NFSP non-convergence in #199. Worth a convergence investigation once cost is tractable.

## Source
#198 cluster calibration, 2026-06-21. #198's perf+observability+checkpointing code is merged; this caching pass is the remaining blocker for the PSRO half of #134.

rjwalters:
Closes #212.

## STEP 1 finding (the key recompute-vs-boundary question)

**Interior payoff-tensor cells are ALREADY cached across outer iterations — they are never recomputed.** This is the "it already only evaluates boundary slabs" branch of the issue, not the "interior cells recomputed" branch.

Trace through `src/multi_agent/psro.rs::run()`:
- `PayoffCache::resize_for_boundary(k -> k+1)` grows storage and **re-keys + preserves** every prior cell (it must rebuild the flat buffer because the little-endian mixed-radix index `s = Σ sᵢ·kⁱ` shifts when the radix grows, but `std::mem::take` moves each old cell into its new slot — payoffs are kept).
- `run()` then collects only the **boundary slab** (joint strategies whose per-agent index vector contains the newest strategy `k-1`) and evaluates just those via `evaluate_payoff_boundary_parallel`. Interior cells are left untouched.
- The pre-existing test `test_payoff_cache_only_evaluates_new_boundary` already pins `eval_count == 1 + K² + 2K` for N=2 — i.e. only the boundary is ever evaluated; the interior is never recomputed.

So a cross-iteration "persistent cache" already exists and works. The dominant remaining cost is the **boundary slab itself**: `(k+1)^N − k^N ≈ N·k^(N-1)` cells, which for the 4-player bucket-brigade game (`N=4`) is `~4·k³` and still super-linear — exactly the asymptotic the #198 calibration measured. Per the issue, the lever in this branch is the `max_population_size` cap (already a hard `Err`) and **subsampling**.

## Fix

Added an opt-in `PsroConfig::max_payoff_evals_per_iteration: Option<usize>`:

- `None` (default) — evaluate the entire boundary slab. **Bit-identical to the pre-#212 behavior.**
- `Some(cap)` with boundary > cap — deterministically subsample `cap` boundary cells via an evenly-spaced stride (`floor(j·len/cap)`, always includes index 0), roll those out through the **existing rayon-parallel path**, and fill each un-sampled cell from its nearest *preceding* sampled neighbour. Filled cells perform no fresh rollout, so `eval_count` continues to count only real evaluations. Selection is a pure function of `(boundary.len(), cap)` — no RNG or thread-order dependence — so the subsampled meta-game is itself fully deterministic.

The hot loop keeps the #201 per-cell-seeded purity and the #203 parallelism untouched; the new `select_boundary_to_evaluate` is a pure pre-pass over the cell-index list.

Example wiring: `MAX_PAYOFF_EVALS_PER_ITER` env var in `examples/games/bucket_brigade/train_psro.rs` (unset = unbounded/bit-identical) so the operator can exercise the cap on-cluster without recompiling.

## Bit-identity test evidence (load-bearing)

`test_subsampling_cap_unreached_is_bit_identical_to_uncapped` runs 3 PSRO iterations on matching pennies from the **same seed** three ways — `None`, `Some(1_000)`, `Some(usize::MAX)` — and asserts the **payoff tensor, per-cell `eval_count`, AND the full exploitability trace are bit-for-bit equal** across all three. PASS. This proves the cache/subsampling plumbing is a no-op whenever the cap is not exceeded, preserving the #201 determinism and #203 parallel bit-identity.

Supporting tests (all PASS):
- `test_select_boundary_to_evaluate` — pure selection/fill correctness (None passthrough, cap≥len passthrough, stratified selection, fill mapping, `Some(0)`→`Some(1)`, determinism).
- `test_subsampling_cap_bounds_evals_and_fills_tensor` — an *exceeded* cap bounds `eval_count` (< the uncapped 16), fully populates the 4×4 tensor (no unfilled cells), and is deterministic across runs.

## Verification

- `cargo test --features training --lib psro` — 30 passed.
- `cargo test --features training --test test_psro_matching_pennies --test test_psro_n_player_matching_pennies --test test_seeded_reproducibility` — all passed (incl. the PSRO exploitability-trace bit-identity reproducibility test).
- `cargo fmt --check` — clean.
- `cargo clippy --all-targets --features training -- -D warnings` — **zero errors in touched files**. Remaining clippy errors are all pre-existing in untouched files (`src/env/games/pong.rs`, `src/env/games/snake/*`, `src/policy/inference.rs`, `src/buffer/rollout/tests.rs`) and are out of scope.
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --features training` — clean.

On-cluster speedup re-measurement (and the recalibrated #134 PSRO budget) is operator-gated.

### DIFF
--- a/examples/games/bucket_brigade/train_psro.rs
+++ b/examples/games/bucket_brigade/train_psro.rs
@@ -203,6 +203,13 @@ fn main() -> Result<()> {
         .ok()
         .and_then(|s| s.parse().ok())
         .unwrap_or(DEFAULT_CHECKPOINT_INTERVAL_ITERATIONS);
+    // Optional per-iteration payoff-eval cap (issue #212). Unset =>
+    // evaluate the full boundary slab (bit-identical to the uncapped
+    // path). Set MAX_PAYOFF_EVALS_PER_ITER=<N> to deterministically
+    // subsample the boundary and bound per-iteration cost on the
+    // 4-player (population⁴) game.
+    let max_payoff_evals_per_iteration: Option<usize> =
+        std::env::var("MAX_PAYOFF_EVALS_PER_ITER").ok().and_then(|s| s.parse().ok());
 
     tracing::info!("Starting PSRO bucket-brigade training (Burn backend: {BACKEND_LABEL})");
     tracing::info!("  cell             = {cell} (β={beta}, κ={kappa}, c={cost})");
@@ -219,6 +226,13 @@ fn main() -> Result<()> {
     tracing::info!("  rollout_steps    = {rollout_steps}");
     tracing::info!("  num_agents       = {NUM_AGENTS}");
     tracing::info!("  hidden_dim       = {HIDDEN_DIM}");
+    tracing::info!(
+        "  max_payoff_evals = {}",
+        match max_payoff_evals_per_iteration {
+            Some(c) => format!("{c}/iter (boundary subsampling on)"),
+            None => "unbounded (full boundary; bit-identical)".to_string(),
+        }
+    );
 
     let device: burn::tensor::Device<InnerBackend> = Default::default();
 
@@ -234,6 +248,7 @@ fn main() -> Result<()> {
         max_population_size: 50,
         br_train_steps_per_iteration: 1,
         payoff_eval_episodes: 1,
+        max_payoff_evals_per_iteration,
         seed: SEED,
     };
     let joint_config = JointTrainerConfig {
--- a/src/multi_agent/psro.rs
+++ b/src/multi_agent/psro.rs
@@ -691,6 +691,79 @@ pub(crate) fn compose_joint_index(components: &[usize], k: usize) -> usize {
     s
 }
 
+/// Decide which boundary cells to actually roll out this iteration, and
+/// how to fill the rest, given an optional per-iteration evaluation cap
+/// (issue #212).
+///
+/// Returns `(to_evaluate, fill_from)`:
+/// - `to_evaluate` is the deterministic subset of `boundary` cells to roll out
+///   (in the *same relative order* as `boundary`, so the downstream rayon
+///   evaluation and cache writes stay deterministic).
+/// - `fill_from` is a list of `(boundary_dst_index, to_evaluate_src_index)`
+///   pairs: boundary cell `boundary[dst]` (which was *not* selected) is to be
+///   filled by copying the payoff of the selected cell evaluated at
+///   `to_evaluate[src]`.
+///
+/// # Selection scheme
+///
+/// - `cap == None`, or `boundary.len() <= cap`: **all** cells are selected and
+///   `fill_from` is empty. This is the default path and is **bit-identical** to
+///   evaluating the whole boundary (the pre-#212 behavior).
+/// - `Some(cap)` with `boundary.len() > cap >= 1`: select `cap` cells by an
+///   evenly-spaced deterministic stride over the boundary index range (`sel_idx
+///   = floor(j * len / cap)` for `j in 0..cap`), guaranteeing a stratified,
+///   reproducible cover that always includes the first cell. Every non-selected
+///   cell is filled from the nearest *preceding* selected cell (the largest
+///   selected index `<= its index`), which is well-defined because index 0 is
+///   always selected. Selection depends only on `(boundary.len(), cap)` — never
+///   on RNG or thread order — so the subsampled meta-game is itself fully
+///   deterministic.
+///
+/// `cap == Some(0)` is treated as `Some(1)` (always roll out at least one
+/// cell) so the boundary is never left entirely unevaluated.
+#[allow(clippy::type_complexity)]
+fn select_boundary_to_evaluate(
+    boundary: &[Vec<usize>],
+    cap: Option<usize>,
+) -> (Vec<Vec<usize>>, Vec<(usize, usize)>) {
+    let len = boundary.len();
+    let cap = match cap {
+        None => return (boundary.to_vec(), Vec::new()),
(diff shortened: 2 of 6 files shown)


rjwalters:
## Judge review: APPROVED

Reviewed locally on `feature/issue-212` @ 85e06b9. The change is additive, opt-in, and the default path is provably a no-op.

### Verification (local; per the ~30min CI policy, local-pass approval)
- `cargo test --features training --lib psro` -> **30 passed, 0 failed** (incl. all three new subsampling tests).
- `cargo test --test test_psro_matching_pennies --test test_psro_n_player_matching_pennies` -> smoke tests pass.
- `cargo fmt --check` clean; `RUSTDOCFLAGS=-D warnings cargo doc` clean; `cargo clippy --lib` -> **zero references to psro.rs** (remaining errors are all pre-existing in untouched files: snake/pong/inference — out of scope).

### Item-by-item
1. **Default `None` == bit-identical to pre-#212**: Confirmed. `test_subsampling_cap_unreached_is_bit_identical_to_uncapped` runs 3 iters three ways (`None` / `Some(1000)` / `Some(usize::MAX)`) and asserts the payoff tensor, per-cell `eval_count`, AND the full exploitability trace are bit-for-bit equal. Code path: `select_boundary_to_evaluate` returns `(boundary.to_vec(), empty)` for `None` or `len <= cap`, so `fill_from` is empty and the fill loop is a no-op. PASS.
2. **Deterministic subsampling**: Confirmed. Selection is `idx = (j*len)/cap` for `j in 0..cap` — pure integer arithmetic over `(boundary.len(), cap)`, always includes index 0, strictly increasing for `len>cap`, no RNG/thread/iteration-order dependence. Fill walks to nearest preceding selected cell. `test_select_boundary_to_evaluate` pins exact outputs and determinism.
3. **Approximation honestly documented**: The authoritative `PsroConfig::max_payoff_evals_per_iteration` doc states the cap "bounds per-iteration cost **at the price of an approximate meta-game on the subsampled boundary**." The `run()` comment and the bit-identity test are both correctly scoped to the *uncapped-equivalent* case ("no-op whenever the cap is not actually exceeded") and do NOT claim capped runs are exact/correct/equivalent. No overclaim. (Non-blocking nit below.)
4. **`eval_count` semantics**: Confirmed. `set_cell_no_count` is `set_cell` minus the `eval_count += 1`, so filled cells (no fresh rollout) are never counted. `test_subsampling_cap_bounds_evals_and_fills_tensor` proves capped `eval_count < 16` (uncapped) while the 4x4 tensor stays fully populated.
5. **Additive/contained**: psro.rs + example + test-config additions only; default behavior unchanged.

### Non-blocking nit (no fix required to merge)
The example knob comment in `examples/games/bucket_brigade/train_psro.rs` ("subsample the boundary and bound per-iteration cost") describes *what* the cap does but omits the approximation tradeoff that the `PsroConfig` doc spells out. Consider adding a one-line "(makes the meta-game approximate — cost/accuracy tradeoff, not a free speedup)" to the operator-facing comment for parity with the API doc. Not blocking since the authoritative field doc is honest and the example never implies exactness.

Approving. `mergeable: MERGEABLE` (no conflicts); CI test jobs were still in-flight at review time (~30min runners) — local verification covers them.

PR #221, opened 2026-06-22
fix: make snake env food respawn deterministic via owned seeded RNG

### LINKED ISSUE #210: Flaky test: snake clone_restore_round_trips intermittently fails on Linux CI
## Symptom
`env::games::snake::environment::tests::clone_restore_round_trips` (`src/env/games/snake/environment.rs:666`) intermittently fails on the **Linux** Tests CI job with `assertion left == right failed` on two `Vec<f32>` (observations), while **passing on macOS** in the same run (e.g. run 27910972424, on PR #209 whose changes are unrelated to the snake env).

Locally it passes 5/5 — so it is a genuine intermittent flake, not a hard regression.

## Likely cause
A `clone_state`/`restore_state` round-trip on the snake env produces observations that don't compare bit-equal under some RNG/iteration-order condition that surfaces on Linux. Candidates: the snake env's reset/step RNG not fully captured in the state snapshot (so restore replays a slightly different stream), or an order-dependent HashMap/iteration in observation construction. The test predates recent work (logic from #70); only #194's unused-import cleanup touched the file since.

## Suggested fix
- Reproduce by running the test in a loop (`for i in $(seq 1 50); do cargo test --features training --lib clone_restore_round_trips; done`), ideally on Linux.
- Ensure `clone_state` captures ALL state that affects subsequent observations (including any RNG), so `restore_state` + step is bit-identical — mirror the deterministic snapshot approach used by `pendulum.rs` / `mountain_car_continuous.rs`.
- If a tiny float nondeterminism is inherent, use a tolerance compare instead of exact `==` (but prefer fixing the determinism).

## Acceptance criteria
- [ ] Root-caused (state-capture gap vs float nondeterminism).
- [ ] Test made deterministic (passes 50/50 on Linux + macOS) without weakening what it checks.

## Source
Surfaced on the #209 CI run (2026-06-21); pre-existing, unrelated to that PR.

rjwalters:
## Summary

`env::games::snake::environment::tests::clone_restore_round_trips` intermittently failed on Linux CI with `assertion left == right failed` on two `Vec<f32>` observations. This PR root-causes and fixes the underlying determinism gap.

## Root cause

The snake env spawned food (initial placement in `new_multi`/`reset`, and respawn in `step_multi`/`step_multi_agents`) using the **thread-local** `rand::rng()`. That RNG was **not** captured by `clone_state`. So after `restore_state`, any step that ate food drew a *different* food cell than the original trajectory — making the resulting observation, reward, and post-step `food` diverge.

The flake only surfaced when the test's experimental `step(0)` happened to land on food (depends on the random initial food position relative to the snake's 6-step path), which is why it failed intermittently and platform-dependently. Reproduced locally at ~1/50 on macOS before the fix; 0/50 after.

This is a state-capture gap (not inherent float nondeterminism), so the fix is to make the snapshot complete rather than loosen the comparison.

## Changes

- Add an owned `rng: StdRng` field to `SnakeEnv` and `SnakeEnvState`, seeded from a new `new_multi_with_seed(width, height, num_agents, seed)` constructor. `new`/`new_multi` delegate with a default seed of `0`.
- Route all food sampling (initial + respawn) through `self.rng` instead of `rand::rng()`.
- Capture/restore the RNG in `clone_state`/`restore_state`, mirroring the deterministic snapshot pattern in `pendulum.rs`.
- Add `clone_restore_round_trips_when_food_is_eaten`, which forces the respawn branch to fire across a restore and asserts the respawned food, observation, and reward all match — so the determinism guarantee is covered regardless of the default seed's trajectory.

| Root-caused (state-capture gap vs float nondeterminism) | done | Confirmed: thread-local RNG used for food respawn not captured in snapshot; reproduced flake locally (~1/50). It is a state-capture gap, not float nondeterminism. |
| Test made deterministic (50/50) without weakening checks | done | Existing exact-`==` assertions kept; added a stronger food-eaten round-trip test. Both pass 50/50 locally (see Test Plan). |

## Test Plan

- `cargo build --features training` — passes.
- `cargo test --features training --lib snake` — 19 passed.
- Loop test (the reproducer): `for i in $(seq 1 50); do cargo test --features training --lib env::games::snake::environment; done` — **50/50 ok** after the fix (was ~1/50 failing before).
- `cargo clippy --features training --lib` — no new warnings in `environment.rs` (the 4 remaining warnings at lines 106/154/514/519 are pre-existing in code untouched by this PR).

Closes #210

### DIFF
--- a/src/env/games/snake/environment.rs
+++ b/src/env/games/snake/environment.rs
@@ -3,7 +3,7 @@
 //! This module implements the SnakeEnv struct and the Environment trait
 //! for both single-agent and multi-agent snake games.
 
-use rand::Rng;
+use rand::{Rng, SeedableRng, rngs::StdRng};
 
 use super::{
     snake::Snake,
@@ -11,13 +11,17 @@ use super::{
 };
 use crate::env::{Environment, SpaceInfo, SpaceType, StepInfo, StepResult};
 
+/// Default seed used when constructing a [`SnakeEnv`] without an explicit seed.
+const DEFAULT_SEED: u64 = 0;
+
 /// Snapshot of a [`SnakeEnv`]'s simulation state.
 ///
 /// Captures every visible field of the env (grid, snake positions, food
-/// position, score, step count, done flag) but **not** the RNG used by
-/// `rand::rng()` inside `step` for food respawn. After `restore_state`, any
-/// step that triggers a food respawn will sample a different food location
-/// than the original trajectory.
+/// position, score, step count, done flag) **and** the env's RNG. Because the
+/// RNG is part of the snapshot, `restore_state` followed by `step` replays the
+/// exact same random stream — so a food respawn after restore spawns food at
+/// the same location as the original trajectory, making round-trips
+/// bit-identical.
 #[derive(Debug, Clone)]
 pub struct SnakeEnvState {
     /// Grid width
@@ -38,6 +42,8 @@ pub struct SnakeEnvState {
     pub max_steps: usize,
     /// Done flag
     pub done: bool,
+    /// RNG state used for food respawn (captured so restores are deterministic)
+    pub rng: StdRng,
 }
 
 /// Multi-agent Snake environment.
@@ -45,11 +51,11 @@ pub struct SnakeEnvState {
 /// # Snapshot semantics
 ///
 /// `clone_state` / `restore_state` capture every visible env field (grid,
-/// snake bodies, food location, score, step count, done flag). They do
-/// **not** capture the thread-local RNG used by `step` to respawn food: if
-/// the restored trajectory triggers a food respawn, the new food spawns at a
-/// different location than in the original. Trajectories that do not eat
-/// food are fully reproducible.
+/// snake bodies, food location, score, step count, done flag) **and** the
+/// env's owned RNG. Because the RNG is part of the snapshot, restoring and
+/// re-stepping replays the identical random stream: a food respawn after
+/// `restore_state` lands in the same cell as the original trajectory, so
+/// round-trips are fully reproducible (including trajectories that eat food).
 #[derive(Debug, Clone)]
 pub struct SnakeEnv {
     /// Grid width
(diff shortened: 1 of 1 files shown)


rjwalters:
✅ **Approved!**

Thorough review of the determinism fix for the flaky `clone_restore_round_trips` test. This is a clean, correctly-scoped fix.

## Verdict on acceptance criteria

**Root cause (state-capture gap vs float nondeterminism)** — Confirmed correct. The `SnakeEnv` food sampling used thread-local `rand::rng()`, which `clone_state` never captured, so a food respawn after `restore_state` drew a different cell and diverged the observation/reward. This is a snapshot-completeness bug, not float nondeterminism. The right fix is to capture state (owned seeded `StdRng`), not to loosen the `==` assertions — and that's what was done.

**Test made deterministic without weakening checks** — Confirmed. The existing exact-equality assertions are unchanged, and a stronger round-trip test was added.

## Verification performed

- **All food-sampling paths route through `self.rng`**: Verified via grep — the only remaining `rand::rng()` in `environment.rs` is inside a doc comment (line 696). Initial placement (`new_multi_with_seed`), `reset`, and both respawn loops (`step_multi` / `step_multi_agents`) now use `self.rng`. The leftover `rand::rng()` calls in `multi_agent.rs` and `snake.rs:239` are out of scope: `multi_agent.rs` is the separate `MultiAgentSnakeEnv` struct (not the type under test), and `snake.rs:239` is inside a `#[test]`.
- **New test genuinely exercises the respawn-across-restore branch**: Verified the mapping — action 0 → `Direction::Up` (delta `(0,-1)`), the snake starts facing Right so Up is accepted (not a reversal), food is placed at `(head.x, head.y-1)` so the next step eats it. The `assert_ne!(food_after_1, food_before)` guards that the respawn branch actually fired; then it asserts food cell, observation, and reward all match across the restore.
- **Reproducibility impact on training**: The training example (`train_snake_multi_v2.rs`) constructs via `new_multi` (now seed 0) and loops `env.reset()`. `reset()` draws from `self.rng` *without re-seeding*, so the stream keeps advancing — food still varies across episodes within a run. The only change is that a run now starts from a fixed seed instead of entropy, i.e. runs become reproducible. This is desirable for an RL framework and mirrors the `pendulum.rs` pattern. No collapse of food to a constant.

## Test execution

- `cargo test --features training --lib snake` — **19 passed**, including the new `clone_restore_round_trips_when_food_is_eaten`.
- Loop reproducer `env::games::snake::environment` x50 — **0/50 failures** (was ~1/50 before the fix).
- `cargo clippy --features training --lib` — the 4 `environment.rs` warnings (lines 106/154/514/519) are pre-existing in untouched code; the diff introduces no new warnings.

## CI / merge state

- `gh pr checks` — all 8 checks pass (Linux + macOS Burn NdArray test jobs, Rustfmt, Docs, Security Audit, etc.).
- `mergeStateStatus` — CLEAN.
- PR description uses `Closes #210`.

Solid root-cause analysis and a well-targeted test. Nice work.

PR #209, opened 2026-06-21
Gate heavy PSRO/NFSP matching-pennies convergence tests behind #[ignore] (closes #208)

rjwalters:
Closes #208

## Summary

The CI `Tests` job (`cargo test --features training`) was ~2h per PR, dominated by the matching-pennies PSRO/NFSP **convergence** training tests. This PR gates those heavy convergence runs behind `#[ignore]`, keeping a fast always-on smoke assertion in each file — matching the repo's existing SAC/A2C convention (`tests/test_a2c_cartpole.rs`, `tests/test_sac_pendulum.rs`).

`test_seeded_reproducibility` is intentionally left in CI (it guards the determinism contract and is cheap).

## What changed (per file)

Each file now has a TINY-budget (2 outer iterations) fast smoke test asserting the trainer runs end-to-end and produces FINITE, structurally-valid outputs (NashConv/exploitability finite & non-negative; per-agent meta-Nash / average-policy action marginals are valid simplex distributions summing to ~1). The full-budget convergence tests are `#[ignore]`d with their **current budgets and assertions kept verbatim**.

| File | Smoke test added | Tests now `#[ignore]`d (verbatim) |
|------|------------------|-----------------------------------|
| `test_psro_n_player_matching_pennies` | `psro_n4_smoke_runs_and_is_finite` | `test_psro_n4_converges_to_uniform_on_majority_game`, `test_psro_n4_nashconv_trend_does_not_blow_up` |
| `test_psro_matching_pennies` | `psro_matching_pennies_smoke_runs_and_is_finite` | `test_psro_converges_to_uniform_on_matching_pennies`, `test_psro_exploitability_non_increasing_trend_on_matching_pennies` |
| `test_nfsp_matching_pennies` | `nfsp_matching_pennies_smoke_runs_and_is_finite` | `test_nfsp_multi_agent_...`, `test_nfsp_single_agent_...`, `test_nfsp_br_oscillates_...` |
| `test_nfsp_n_player_matching_pennies` | `nfsp_n4_smoke_runs_and_is_finite` | `test_nfsp_n4_converges_to_uniform_on_majority_game`, `test_nfsp_n4_eta_mixing_rate_in_aggregate` |

For the N-player PSRO file the shared trainer wiring was factored into a `run_psro_n4(max_iterations, seed)` helper so only the iteration budget differs between the smoke (2-iter, `3^4=81`-cell payoff cache) and convergence (5-iter, `6^4=1296`-cell) tiers.

[...]

Issue #208, opened 2026-06-21
CI Tests job is ~2h — gate the heavy PSRO/NFSP matching-pennies training tests


rjwalters:
## Problem

Every PR is gated ~1.5–2h on the `Tests (… Burn NdArray)` CI jobs (`cargo test --features training`, Linux + macOS). Measured per-test-binary timing from a completed run (#206's Linux run, 2026-06-21):

| Test binary | non-ignored tests | wall-clock |
|-------------|-------------------|-----------|
| **`test_psro_n_player_matching_pennies`** | 2 | **~68 min (Linux) / ~61 min (macOS)** |
| lib unit tests | 377 | ~18–23 min |
| `test_psro_matching_pennies` | 2 | ~6 min |
| `test_nfsp_*_matching_pennies` | 2 each | ~3–5 min each |
| `test_seeded_reproducibility` | 2 | ~3 min |

`test_psro_n_player_matching_pennies` alone is **~55% of the entire Tests job** — it runs a full N-player PSRO convergence training (the α-rank `population⁴` payoff tensor) on every PR.

The convergence tests for the *other* algorithms (SAC/A2C/BC/DQN/bucket-brigade) are already `#[ignore]`d in CI with fast smoke tests kept always-on. The matching-pennies PSRO/NFSP tests are the inconsistent holdouts that run full training per-PR.

## Fix (apply the repo's existing convention)

Make the matching-pennies PSRO/NFSP tests consistent with the rest of the suite:

1. **`test_psro_n_player_matching_pennies`** (the dominant cost): split into a **fast always-on smoke test** (tiny budget — e.g. 2–3 outer iterations — asserting the run executes and produces finite/valid exploitability/NashConv values) + an **`#[ignore]`d full-budget convergence test** (the current trend/NashConv assertions). This removes ~68 min from CI.
2. Review `test_psro_matching_pennies` / `test_nfsp_*_matching_pennies` similarly; keep them in CI if they're already <~2 min after #207's parallelization, otherwise apply the same smoke + `#[ignore]` split.
3. Keep `test_seeded_reproducibility` in CI (it's ~3 min and guards the determinism contract).

Net: per-PR CI drops from ~2h to ~20–30 min, while the full convergence assertions remain runnable on demand (`cargo test … -- --ignored`) / nightly.

Note: #207 (merged) parallelized the PSRO payoff tensor, which already cuts the N-player test substantially on multi-core hosts — this issue makes the per-PR CI cost bounded regardless.

[...]

PR #238, opened 2026-06-23
perf(psro): parallelize round-robin best-response loop with rayon

### LINKED ISSUE #232: PSRO: parallelize the round-robin best-response loop (~4x, reuse #207 rayon pattern)
## Problem

NFSP/PSRO bucket-brigade training on a 32-core box runs at load ~1.0 — effectively single-threaded. For PSRO the dominant *parallelizable* axis is the round-robin best-response loop, which trains `num_agents` (=4) **fully independent** best responses serially.

## Evidence (from parallelism investigation, 2026-06-23)

- The BR loop `for active_agent in 0..num_agents` is at `src/multi_agent/psro.rs:1591-1594`.
- Each `train_best_response` (`psro.rs:1875-1948`) builds its **own** `JointMultiAgentTrainer`, its **own** env (`psro.rs:1909`), and only reads `self.populations` / `self.config`.
- The only shared-mutable touches are `self.rng` (opponent sampling `psro.rs:1893`, PPO shuffle `psro.rs:1916,1936`) and `self.next_init_seed` (`psro.rs:1890`) — both pre-computable before a parallel region.
- PSRO payoff eval is **already** parallelized with exactly the right pattern (#207): `evaluate_payoff_boundary_parallel` (`psro.rs:2007-2040`) binds `Sync` locals, delegates to a pure locally-seeded fn, and uses `par_iter().map(...).collect()` so results are **order-preserved / bit-identical regardless of thread count**. `rayon` is already a dep (`Cargo.toml:72`).
- Backend tensor types are confirmed `Send + Sync` (the payoff path already requires `P: Send + Sync, E: Send, FE: Sync, B::Device: Sync` at `psro.rs:1557-1560` and runs `Autodiff<NdArray<f32>>` under `par_iter`).

## Fix

Parallelize the round-robin BR loop using the #207 pattern:
1. Pre-draw per-agent opponent indices + init seeds from `self.rng` / `next_init_seed` **before** the parallel region (fixed order).
2. Give each BR a **local `StdRng`** seeded deterministically from `(config.seed, active_agent)` (mirrors `evaluate_payoff_joint_pure` at `psro.rs:2105-2106`).
3. Refactor `train_best_response` to not touch `&mut self` / `&mut self.rng` (pure-ish), then `(0..num_agents).into_par_iter().map(...).collect()` and append to `self.populations` in **fixed agent order** after the join (collect-by-index, like `psro.rs:1645-1651`).

## Acceptance criteria

- [ ] The BR loop runs `num_agents` best responses concurrently (load > 1 during the BR phase on a multicore box).
- [ ] Results are deterministic across thread counts (per-agent local seeded RNG; fixed-order population append). Document that bit-identity vs the *old* serial-RNG runs is intentionally not preserved (RNG threading changed), but runs are reproducible for a given seed/thread-invariant.
- [ ] Existing PSRO tests pass; add a test asserting thread-count-invariant output (mirror the payoff-eval determinism test).

## Expected impact

~`num_agents`× (≈4×) on the BR-training phase — the bulk of PSRO per-iteration cost now that payoff eval is already parallel. (Scales with `num_agents`, not core count; further rollout-level parallelism is a separate issue.)

## Source

Parallelism investigation during the #134 distributed re-run (2026-06-23). Does NOT affect the in-flight runs (they use already-built binaries); benefits future re-runs.

rjwalters:
## Summary

Parallelize the PSRO round-robin best-response loop. Previously the `num_agents` (=4) **fully-independent** best responses were trained serially (`for active_agent in 0..num_agents`), leaving a 32-core box at load ~1.0. They now run concurrently under rayon, reusing the exact #207 payoff-eval pattern (`evaluate_payoff_boundary_parallel` + the pure, locally-seeded `evaluate_payoff_joint_pure`).

This is a code change for future runs only — it does not affect any in-flight cluster jobs (they use already-built binaries).

## Approach

- **Hoist all shared-mutable draws out of the parallel region.** Before the join, in fixed agent order, draw each agent's frozen opponent indices from `self.rng` and each active BR's init seed from `self.next_init_seed()`. The parallel region then touches no `&mut self` / `&mut self.rng`.
- **Extract a pure per-task body.** The old `train_best_response` is refactored into a free function `train_best_response_pure` that takes shared `&` references plus the pre-drawn inputs and returns `(JointStats, P)` — no `&mut self`. Each BR runs under a **local `StdRng` seeded from `(config.seed, active_agent)`** (via `splitmix64`), mirroring `evaluate_payoff_joint_pure`'s per-cell seeding.
- **Collect-by-index join.** The loop becomes `(0..num_agents).into_par_iter().map(...).collect()`; results are unpacked in fixed agent order and trained policies appended to `self.populations` in that same order after the join, so the population layout is independent of scheduling.
- **Bounds.** `run` / `run_silent` / `train_best_responses_parallel` add `FP: Sync, FO: Sync` (each task calls the policy/optimizer factories through a shared `&`) on top of the existing `P: Send + Sync, E: Send, FE: Sync, B::Device: Sync` style from the payoff path.

## Determinism guarantee

- **Thread-count-invariant:** for a fixed seed, the per-agent populations produced by a full PSRO run are **byte-identical** whether the BR loop runs under a 1-thread or a 4-thread rayon pool. This is guaranteed by drawing opponent indices/init seeds in fixed order before the parallel region and giving each BR its own deterministic local RNG.
- **Not bit-identical to the old serial-RNG runs (by design, documented).** Each BR now consumes its own local RNG stream instead of slices of the single shared `self.rng` stream, so output differs from pre-#232 runs. It remains fully reproducible for a given seed.

## Changes

- `src/multi_agent/psro.rs`:
  - Replace serial BR loop in `run` with `train_best_responses_parallel`.
  - New method `train_best_responses_parallel` (fixed-order pre-pass + `into_par_iter` + fixed-order append).
  - New pure free function `train_best_response_pure` (no `&mut self`, local per-agent RNG).
  - Add `FP: Sync, FO: Sync` bounds to `run` / `run_silent`.
  - New test `test_best_response_parallel_thread_count_invariant`.

| BR loop runs `num_agents` BRs concurrently | ✅ | `(0..num_agents).into_par_iter().map(train_best_response_pure)` |
| Deterministic across thread counts | ✅ | New test asserts byte-identical populations under 1- vs 4-thread rayon pools (3-iteration run) |
| Bit-identity vs old serial-RNG documented as intentionally dropped | ✅ | Documented in fn docs, commit, and here |
| Existing PSRO tests pass | ✅ | `cargo test --features training --lib psro` → 34 passed; `test_psro_matching_pennies` smoke → ok |
| New determinism test mirrors payoff-eval test | ✅ | `test_best_response_parallel_thread_count_invariant` |

## Test Plan

- `cargo build --features training` → ok
- `cargo test --features training --lib psro` → 34 passed; 0 failed (incl. new test)
- `cargo test --features training --test test_psro_matching_pennies` → smoke ok (convergence tests `#[ignore]`d as before)
- `cargo fmt -- --check` → clean
- `cargo clippy --all-targets --features training -- -D warnings` → clean
- `cargo clippy --all-targets --features "training,env-bucket-brigade" -- -D warnings` → clean

Closes #232

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,13 +78,7 @@ jobs:
           submodules: recursive
       - uses: dtolnay/rust-toolchain@nightly
       - name: cargo test --features training
-        # `--test-threads=1` avoids rayon oversubscription on the 2-core CI
-        # runners: the #232 PSRO best-response loop (and other `par_iter`
-        # paths) dispatch heavy work to the global rayon pool, so running
-        # multiple test fns concurrently would oversubscribe the 2 cores and
-        # roughly double the suite wall-clock. Serial test fns each get the
-        # full pool, keeping the suite near its pre-#232 baseline.
-        run: cargo test --features training -- --test-threads=1
+        run: cargo test --features training
 
   # Same on macOS, to catch any platform-specific Burn issues early.
   test_macos:
@@ -99,13 +93,7 @@ jobs:
         run: cargo check --features training
       # Run tests as a separate step so a hang doesn't mask the check result.
       - name: cargo test --features training
-        # `--test-threads=1` avoids rayon oversubscription on the 2-core CI
-        # runners: the #232 PSRO best-response loop (and other `par_iter`
-        # paths) dispatch heavy work to the global rayon pool, so running
-        # multiple test fns concurrently would oversubscribe the 2 cores and
-        # roughly double the suite wall-clock. Serial test fns each get the
-        # full pool, keeping the suite near its pre-#232 baseline.
-        run: cargo test --features training -- --test-threads=1
+        run: cargo test --features training
 
   # Documentation
   docs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,7 +78,13 @@ jobs:
           submodules: recursive
       - uses: dtolnay/rust-toolchain@nightly
       - name: cargo test --features training
-        run: cargo test --features training
+        # `--test-threads=1` avoids rayon oversubscription on the 2-core CI
+        # runners: the #232 PSRO best-response loop (and other `par_iter`
+        # paths) dispatch heavy work to the global rayon pool, so running
+        # multiple test fns concurrently would oversubscribe the 2 cores and
+        # roughly double the suite wall-clock. Serial test fns each get the
+        # full pool, keeping the suite near its pre-#232 baseline.
+        run: cargo test --features training -- --test-threads=1
 
   # Same on macOS, to catch any platform-specific Burn issues early.
   test_macos:
@@ -93,7 +99,13 @@ jobs:
         run: cargo check --features training
       # Run tests as a separate step so a hang doesn't mask the check result.
       - name: cargo test --features training
-        run: cargo test --features training
+        # `--test-threads=1` avoids rayon oversubscription on the 2-core CI
+        # runners: the #232 PSRO best-response loop (and other `par_iter`
+        # paths) dispatch heavy work to the global rayon pool, so running
+        # multiple test fns concurrently would oversubscribe the 2 cores and
+        # roughly double the suite wall-clock. Serial test fns each get the
+        # full pool, keeping the suite near its pre-#232 baseline.
+        run: cargo test --features training -- --test-threads=1
 
   # Documentation
   docs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,6 +78,16 @@ jobs:
           submodules: recursive
       - uses: dtolnay/rust-toolchain@nightly
       - name: cargo test --features training
+        # Pin rayon to a single worker for the test run only. The #232 PSRO
+        # best-response loop (and other `par_iter` paths) dispatch heavy work
+        # to the global rayon pool; on the 2-core CI runners, multiple rayon
+        # workers spin-contend with cargo's concurrent test threads on the
+        # saturated cores, which ~4x'd the Tests-job wall-clock (11min ->
+        # 40min+). One rayon worker => the par_iter paths run serially in CI
+        # (no spin-contention) while cargo keeps its normal test concurrency.
+        # Real runs (cluster, all cores) are unaffected — they don't set this.
+        env:
+          RAYON_NUM_THREADS: "1"
         run: cargo test --features training
 
   # Same on macOS, to catch any platform-specific Burn issues early.
@@ -93,6 +103,16 @@ jobs:
         run: cargo check --features training
       # Run tests as a separate step so a hang doesn't mask the check result.
       - name: cargo test --features training
+        # Pin rayon to a single worker for the test run only. The #232 PSRO
+        # best-response loop (and other `par_iter` paths) dispatch heavy work
+        # to the global rayon pool; on the 2-core CI runners, multiple rayon
+        # workers spin-contend with cargo's concurrent test threads on the
+        # saturated cores, which ~4x'd the Tests-job wall-clock (11min ->
+        # 40min+). One rayon worker => the par_iter paths run serially in CI
+        # (no spin-contention) while cargo keeps its normal test concurrency.
+        # Real runs (cluster, all cores) are unaffected — they don't set this.
+        env:
+          RAYON_NUM_THREADS: "1"
         run: cargo test --features training
 
   # Documentation
--- a/src/multi_agent/psro.rs
+++ b/src/multi_agent/psro.rs
@@ -3533,7 +3533,8 @@ mod tests {
     /// 2-core CI runners, which hung the Tests job (#232 review). This keeps
     /// cheap, deterministic always-on coverage of the parallel path; the
     /// cross-thread-count (1 vs 4) invariance proof lives in the
-    /// `#[ignore]`d `test_best_response_parallel_thread_count_invariant_thorough`.
+    /// `#[ignore]`d
+    /// `test_best_response_parallel_thread_count_invariant_thorough`.
     ///
     /// Each BR draws its opponent indices + init seed in fixed agent order
     /// before the parallel region and runs under a per-agent local RNG
@@ -3552,10 +3553,7 @@ mod tests {
             "expected populations to grow over the iterations (got {})",
             a[0].len()
         );
-        assert_eq!(
-            a, b,
-            "PSRO best-response output must be deterministic for a fixed seed"
-        );
+        assert_eq!(a, b, "PSRO best-response output must be deterministic for a fixed seed");
     }
 
     /// Thorough multi-iteration variant of the thread-count-invariance
--- a/src/multi_agent/psro.rs
+++ b/src/multi_agent/psro.rs
@@ -3439,82 +3439,130 @@ mod tests {
         }
     }
 
+    /// Run a multi-iteration PSRO trainer (so the parallel BR loop
+    /// executes several times) under a rayon pool of `threads` threads,
+    /// and return the flattened per-agent population policy weights.
+    ///
+    /// `max_iterations` / `rollout_steps` / `hidden` / `br_train_steps` /
+    /// `payoff_eval_episodes` are parameters so callers can pick a tiny
+    /// always-on smoke workload or a heavier `#[ignore]`d proof. Both run
+    /// the same code path (the #232 par_iter BR loop with `num_agents > 1`).
+    #[cfg(test)]
+    fn psro_populations_under_threads(
+        threads: usize,
+        max_iterations: usize,
+        rollout_steps: usize,
+        hidden: usize,
+        br_train_steps: usize,
+        payoff_eval_episodes: usize,
+    ) -> Vec<Vec<Vec<f32>>> {
+        let device: NdArrayDevice = Default::default();
+        let psro_config = PsroConfig {
+            max_iterations,
+            max_population_size: 50,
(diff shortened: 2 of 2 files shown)


rjwalters:
❌ **Changes Requested — CI-time regression from the new determinism test**

The parallelization itself is **correct and well-built** — I verified all the load-bearing correctness points (details below). But the new test makes the always-on CI `Tests` job ~6× slower, which blocks approval.

## The blocker: `test_best_response_parallel_thread_count_invariant` is far too heavy for the CI suite

**Measured CI-time regression:**
- `Tests (Linux + Burn NdArray)` on `main`: **~11 min** (latest run `28004151145`: 05:21:26 → 05:32:45).
- `Tests (macOS + Burn NdArray)` on `main`: **~8 min**.
- This PR's run (`28004971570`): both `Tests` jobs have been **in progress for 65+ minutes** and have not finished. CI never went green; merge state is `UNSTABLE`.

**Root cause:** the test runs a *full 3-iteration PSRO trainer twice* (once under a 1-thread rayon pool, once under 4-thread), each iteration doing PPO rollouts + updates + the payoff sweep. Timed in isolation on a fast M-series Mac with pre-built artifacts it takes **85.8 s** for that one test. On CI's slower shared runners it is multiple minutes, and the 1-thread `pool.install(...)` half is maximally slow because *both* the BR loop and the payoff eval collapse onto a single thread. (`cargo test --features training` runs in plain debug, no `--release`, which compounds it.)

This is a load-bearing build/CI-time issue, not advisory: the job went from ~11 min to >65 min and didn't complete.

### What to fix
Keep the thread-count-invariance assertion (it's the right test) but make it cheap enough for the always-on suite. Options, in order of preference:

1. **Shrink the workload drastically.** This test does not need real convergence — only byte-identical output across thread counts. Cut `max_iterations` to ~2, `br_train_steps_per_iteration` to 1, `rollout_steps` to ~8, `payoff_eval_episodes` to 1-2, and hidden size to ~4-8. The invariance property holds regardless of workload size, so a sub-second run still proves it.
2. **And/or gate the heavy variant behind `#[ignore]`** (mirroring #209's treatment of the matching-pennies convergence tests), leaving only a tiny always-on smoke version of the invariance check.

Target: the new test should add only a second or two to the suite so the `Tests` job stays near the ~11 min `main` baseline.

## Correctness review (all passed — no changes needed here)

1. **No shared-mutable aliasing in the parallel region** ✅ — `train_best_response_pure` takes no `&mut self`; the closure captures only immutable locals (`populations`, `config`, `joint_config`, `device`, the three factories) plus pre-drawn `opp_indices`/`init_seeds`, and a *local* `StdRng`. Every `self.rng` draw (`sample_from_mixture`) and `self.next_init_seed()` happens in the fixed-order pre-pass *before* `into_par_iter()` (psro.rs:1946-1957). Nothing in the parallel closure touches `self.rng` or mutates state.
2. **Determinism is order-stable** ✅ — `into_par_iter().collect()` preserves input order; results are unpacked by index and appended to `self.populations[active_agent]` in fixed agent order after the join (psro.rs:2004-2006), independent of completion order. Error propagation is deterministic (first error in index order). Per-agent local seed `config.seed ^ splitmix64(active_agent)` is independent of scheduling and mirrors `evaluate_payoff_joint_pure` exactly.
3. **Opponent sampling semantics preserved** ✅ — opponents are sampled from `per_agent_marginals` (the pre-iteration meta-Nash mixture) into the pre-iteration `populations` snapshot. The old serial loop already could never select a same-iteration freshly-appended BR (the marginal vector length covers only the pre-iteration population), so reading the immutable snapshot in parallel is semantically identical.
4. **Trait bounds** ✅ — `FP: Sync, FO: Sync` are sound (factories invoked via shared `&`) and don't over-constrain real callers: `clippy --all-targets --features "training,env-bucket-brigade"` passes clean, so the bucket-brigade `train_psro` example still compiles.

## Local checks (all green)
- `cargo clippy --all-targets --features training -- -D warnings` → clean
- `cargo clippy --all-targets --features "training,env-bucket-brigade" -- -D warnings` → clean
- `cargo fmt -- --check` → clean
- `cargo test --features training --lib psro` → 34 passed (incl. the new test, but it dominates at ~86 s)
- `cargo test --features training --test test_psro_matching_pennies` → smoke ok; convergence tests `#[ignore]`d

Once the new test is lightened so the `Tests` job returns to roughly its `main` baseline (and CI goes green), I'll re-approve. Everything else is ready.

rjwalters:
Addressed the single review item: the new thread-count-invariance test was too slow for CI.

**Fix (test-only — production parallelization code untouched):**
- `test_best_response_parallel_thread_count_invariant` is now an always-on smoke at a tiny workload (2 iterations, 1 BR train step, 8 rollout steps, hidden=4, 1 payoff episode). It still drives the real #232 `par_iter` BR loop with `num_agents=2` (so the loop runs concurrently under 4 threads) and keeps the load-bearing assertion: **byte-identical per-agent populations under a 1-thread vs 4-thread rayon pool**.
- Runtime dropped from ~85.8s to **5.29s** in the default debug test lane.
- The heavier 3-iteration proof moved to a new `#[ignore]`d variant `test_best_response_parallel_thread_count_invariant_thorough` (opt in with `--ignored`, prefer `--release`), matching the #208/#209 convention.

`train_best_responses_parallel` / `train_best_response_pure` are unchanged (diff is confined to the `#[cfg(test)]` module in `src/multi_agent/psro.rs`).

Verified: `cargo fmt -- --check` clean, `cargo clippy --all-targets --features training -- -D warnings` clean, the smoke test passes and the thorough test is correctly ignored.

Flipping back to `loom:review-requested`.

rjwalters:
❌ **Changes Requested — CI `Tests` jobs do not complete (reproducible across two runs)**

Thanks for lightening the test — that part is correct and the new always-on smoke is genuinely cheap in isolation (I measured **0.45s** under `RAYON_NUM_THREADS=2`). But the PR still cannot be approved because **the CI `Tests` jobs never go green**, and this reproduced across two independent runs.

## Evidence

| Run | Tests (Linux) | Tests (macOS) | Outcome |
|-----|---------------|----------------|---------|
| `28008025954` (1st attempt) | **51m32s** | 51m33s | never finished (I cancelled) |
| `28008025954` (re-run) | **18m+** | 18m+ | never finished (I cancelled) |
| Peer PRs today (issue-233/234/229, main) | 8–13 min | 8–11 min | ✅ pass |

The `cargo test --features training` **step itself** ran 51 min then 18 min on the two attempts (compile is identical to main, so the delta is test *runtime*). Other PRs and `main` ran the same step in ~11 min today, so this is not a runner-wide outage — it is specific to this branch.

## What I could and couldn't reproduce locally

- Full lib suite, normal: **101s**.
- Full lib suite, `RAYON_NUM_THREADS=2` + default (parallel) test threads — i.e. simulating a 2-core CI runner: **253s, exit 0, no hang.**
- The new test alone under 2-core limit: **0.45s.**

So I cannot reproduce a hang on my (many-core) box, yet CI reproducibly balloons. The most likely cause is **rayon thread-pool contention on low-core CI runners**: this PR adds another `ThreadPoolBuilder::new().num_threads(N).build()` + `pool.install(|| … into_par_iter() …)` site (the new invariance test) on top of the existing payoff-parallel test. When `cargo test` runs many tests concurrently on a 2-core runner, nested/competing rayon pools can serialize or thrash badly in a way that simply doesn't show on a high-core dev box.

## Requested fix (pick the one that makes CI green; #1 is the most likely root-cause fix)

1. **Avoid creating bespoke multi-thread rayon pools inside an always-on test.** The invariance property only needs *different effective thread counts*, not two freshly-built pools competing with `cargo test`'s own parallelism. Options:
   - Run the always-on smoke comparing `num_threads(1)` vs `num_threads(2)` (not 4), and/or mark it `#[serial]` (via `serial_test`) so it doesn't overlap the payoff-parallel test on a 2-core box.
   - Or move the 1-thread/4-thread comparison entirely into the `#[ignore]`d `…_thorough` variant, and have the always-on smoke assert invariance a cheaper way (e.g. two runs under the *same* default pool must match — still catches the ordering/seed bug, costs ~nothing).
2. **And/or pin CI test threads** so concurrent rayon-pool tests can't oversubscribe: change the two `cargo test --features training` steps in `.github/workflows/ci.yml` (lines 81 and 96) to `cargo test --features training -- --test-threads=1` (or set `RUST_TEST_THREADS`). One-line infra change that deterministically removes the cross-test rayon contention.

**Acceptance:** CI `Tests (Linux)` and `Tests (macOS)` must complete green in roughly the ~11 min `main` baseline. Please push the fix and re-request review; I'll confirm the timing on the fresh run before approving.

## Everything else is ready (re-verified on this commit)

The production parallel-BR code (`train_best_responses_parallel` / `train_best_response_pure`) is unchanged and correct, and the test refactor itself is clean:
- No `&mut self` / shared-RNG in the parallel region; opponent indices + init seeds drawn in the fixed-order pre-pass; collect-by-index append in fixed agent order. ✅
- `cargo clippy --all-targets --features training -- -D warnings` → clean ✅
- `cargo clippy --all-targets --features "training,env-bucket-brigade" -- -D warnings` → clean ✅
- `cargo fmt -- --check` → clean ✅
- `cargo test --features training --lib psro` → 34 passed, 1 ignored (the new `…_thorough`) ✅
- The lightened smoke is 1.2s warm / 0.45s under 2-core limits ✅

The only blocker is getting the CI `Tests` jobs to actually finish green.

Issue #239, opened 2026-06-23
PPO best-response does not learn on bucket-brigade (root cause of #199/#215 non-convergence): value gradient swamps policy on shared trunk + dead grad-clip + BR undertraining


rjwalters:
## Summary

The PPO **best-response (BR) oracle does not learn on the bucket-brigade cells**, and this is the shared root cause of both the NFSP non-convergence (#199) and the PSRO non-convergence (#215). Surfaced empirically in the #134 full-scale re-run (48×8192, all 3 cells, 2026-06-22/23): with the #199/#215 knobs active (`AP_COVERAGE=2.0`, `BR_REWARD_SCALE=0.01`, `ALPHA_RANK_NORMALIZE_SPAN=1`), the BR stays at uniform-random and neither trainer converges. The micro-scale gains from #199/#215 did **not** generalize to the real budget.

## Empirical evidence (live #134 runs)

NFSP per-iteration BR diagnostics (cell beta05, iters 1–4):

| iter | avg_ap_loss | br pol | br val | br ent |
|------|-------------|--------|--------|--------|
| 1 | 3.6857 | −0.033 | 910.9 | 1.2295 |
| 2 | 3.6851 | −0.003 | 880.5 | 1.2292 |
| 3 | 3.6852 | +0.011 | 989.5 | 1.2290 |
| 4 | 3.6852 | −0.010 | 925.5 | 1.2287 |

- `avg_ap_loss` pinned at `ln(40)=3.689` (uniform floor) — reproduces #199 at scale.
- BR policy loss ≈ 0 and oscillating (no movement); BR value loss **flat at ~900** (critic never fits); BR entropy **pinned at exactly the 1.2297 per-dim uniform-max floor** (dead policy gradient).
- PSRO exploitability decreased for 3 iters (8356→7489→6660 — the #215 micro pattern) then **diverged** (→12840, oscillating 8k–14k) as the population filled with weak BRs — reproduces #215 at scale.

## Root-cause mechanism

The BR is a **shared-trunk actor-critic**: `fc1`/`fc2` are shared by the policy heads and the value head (`src/policy/multi_discrete_mlp.rs:30-34`, `forward` `:162-167`), trained with **one optimizer and one combined backward** `policy_loss + 0.5·value_loss − 0.01·entropy` (`src/multi_agent/joint.rs:783-821`).

[...]

PR #240, opened 2026-06-24
fix(joint): apply grad-norm clip + iterate all minibatches in BR update

### LINKED ISSUE #4: multi_agent::PolicyLearner has unresolved TODO placeholders

rjwalters:
Part of #239 — lands the BR/PPO fix infrastructure and converts the issue's H1 hypothesis into an evidence-backed conclusion. **The BR-learns acceptance criterion is NOT yet met** (see "Honest finding"); #239 stays open for the deeper critic-fitting fix.

## What changed

- **`src/multi_agent/joint.rs`**
  - Added `clip_grads_by_global_norm` — a Burn module-visitor computing the global L2 norm across a policy's gradient tensors and scaling them down to the cap (direction-preserving), fetching/re-registering on the inner (non-autodiff) backend (gradients live on `B::InnerBackend`, not `B`). Wired into the per-policy step so the previously-dead `JointTrainerConfig::max_grad_norm` now actually applies.
  - Added `JointTrainerConfig::iterate_all_minibatches` (default `false` → existing NFSP/PSRO determinism tests bit-identical; `true` → walks every minibatch per epoch).
  - Two new unit tests.
- **`src/train/optimizer.rs`** — corrected stale doc.
- **`examples/games/bucket_brigade/train_nfsp.rs` / `train_psro.rs`** — exposed `VF_COEF` (default 0.05) and `BR_TRAIN_STEPS` env overrides; opted both into `iterate_all_minibatches`.

## A/B validation — NFSP, beta05, 6 iters

| arm | br val (iter1 to 6) | br ent (iter6) | avg_ap_loss Δfloor (iter6) |
|---|---|---|---|
| BASELINE (origin/main, scale=0.01) | 888 to 868 | 1.2288 | -0.0110 |
| FIX (clip + all-mb + vf=0.05, scale=0.01) | 878 to 867 | 1.2279 | -0.0102 |
| FIX + scale=0.001 (3 iters) | 8.1 to 8.1 | 1.2286 | -0.0078 |

## Honest finding (the key result)

The fix does NOT move entropy below 1.0 — the BR still does not learn. But the evidence is decisive: `BR_REWARD_SCALE=0.001` collapses BR value loss from ~900 to ~8, so the value gradient can no longer swamp the shared trunk — yet entropy stays pinned at the uniform-max floor (~1.228) and `avg_ap_loss` stays at `ln(40)`. This rules out gradient magnitude (H1) as the binding constraint. The low-scale value loss is small but flat (8.08, 7.60, 8.11): the critic isn't fitting. The binding constraint is critic-not-fitting / non-stationary opponent → issue #239 fix #4 (separate critic optimizer / higher value-head LR) plus a deeper look (partial observability, return/GAE correctness, network capacity).

This PR lands the correct, tested infrastructure (working grad-clip, all-minibatch iteration, knobs) that any follow-up fix needs.

## Gates

`cargo test --lib multi_agent` = 69 passed, 0 failed, 3 ignored; clippy clean on both `training` and `training,env-bucket-brigade`; `cargo +nightly fmt --check` clean; release build with `env-bucket-brigade` builds.

### DIFF
--- a/examples/games/bucket_brigade/train_nfsp.rs
+++ b/examples/games/bucket_brigade/train_nfsp.rs
@@ -215,13 +215,29 @@ fn main() -> Result<()> {
         .ok()
         .and_then(|s| s.parse().ok())
         .unwrap_or(0.01);
+    // Issue #239: the BR oracle did not learn on bucket-brigade. Root cause
+    // was the raw-scale value-loss gradient swamping the policy heads on the
+    // shared actor-critic trunk, compounded by dead grad-clip config and BR
+    // undertraining (one 256-sample minibatch per epoch). The knobs below
+    // address that; all are env-overridable so the A/B harness in #239 can
+    // sweep them.
+    //   * `VF_COEF` lowers the value-loss weight (0.5 -> 0.05) so the critic
+    //     gradient no longer dominates the shared trunk.
+    //   * `BR_TRAIN_STEPS` runs the PPO update more than once per iteration.
+    //   * grad-clip (`max_grad_norm`, default 0.5) is now actually applied in the
+    //     joint update, and the update iterates ALL minibatches per epoch.
+    let vf_coef: f64 = std::env::var("VF_COEF").ok().and_then(|s| s.parse().ok()).unwrap_or(0.05);
+    let br_train_steps: usize =
+        std::env::var("BR_TRAIN_STEPS").ok().and_then(|s| s.parse().ok()).unwrap_or(1);
     tracing::info!("  ap_coverage      = {ap_coverage} (issue #199 adaptive AP-step floor)");
     tracing::info!("  br_reward_scale  = {br_reward_scale} (issue #199 payoff rescale)");
+    tracing::info!("  vf_coef          = {vf_coef} (issue #239 value-loss weight)");
+    tracing::info!("  br_train_steps   = {br_train_steps} (issue #239 BR updates/iter)");
     let nfsp_config = NfspConfig {
         max_iterations: total_iterations,
         anticipatory_param: 0.1,
         reservoir_capacity: 16_384,
-        br_train_steps_per_iteration: 1,
+        br_train_steps_per_iteration: br_train_steps,
         avg_policy_train_steps_per_iteration: 8,
         avg_policy_minibatch_size: 64,
         avg_policy_lr: 5e-3,
@@ -234,6 +250,9 @@ fn main() -> Result<()> {
         rollout_steps,
         n_epochs: 4,
         minibatch_size: 256,
+        vf_coef,
+        // Issue #239: consume the full rollout instead of one 256-sample draw.
+        iterate_all_minibatches: true,
         ..Default::default()
     };
 
--- a/examples/games/bucket_brigade/train_psro.rs
+++ b/examples/games/bucket_brigade/train_psro.rs
@@ -240,6 +240,13 @@ fn main() -> Result<()> {
         .ok()
         .map(|s| s == "1" || s.eq_ignore_ascii_case("true"))
         .unwrap_or(false);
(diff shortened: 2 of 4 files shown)


rjwalters:
✅ **Approved!**

Evaluated as **infrastructure + decisive negative result** ("Part of #239", not "Closes") — judged on correctness, regression-safety, and evidence soundness, not on whether the BR learns (the PR honestly reports it does not, correctly scoped to a #239 follow-up).

## Load-bearing checks

**1. Regression-safe defaults — PASS.** `iterate_all_minibatches` defaults to `false`. The default branch is bit-identical to the historical path (shuffle → truncate → single `vec![indices]`), and in that path `num_mb_steps == n_epochs`, so the new averaging divisor `n` exactly preserves the prior epoch-averaging. All 69 `multi_agent` lib tests pass locally (0 failed, 3 ignored), including the bit-identical/determinism guards (`test_nfsp_determinism_within_module_under_seed`, `test_payoff_boundary_parallel_matches_serial_bit_identically`, `test_subsampling_cap_unreached_is_bit_identical_to_uncapped`). The library `JointTrainerConfig` `vf_coef` default is unchanged (0.5); only the *examples* opt into 0.05 — correct (examples are not library defaults). `max_grad_norm` default 0.5 was already staged on every optimizer and is now actually applied; existing tests still pass under it.

**2. Grad-clip correctness — PASS.** `clip_grads_by_global_norm` computes the global L2 norm across ALL of a policy's grad tensors (visitor pass 1: `Σ‖g‖²` → sqrt), scales only when `total_norm > max_norm`, direction-preservingly via `coef = max_norm/(total_norm+1e-6)`. Grads are correctly fetched/re-registered on `B::InnerBackend` (where Burn stores them, not the autodiff backend) — the subtle part is handled. Non-finite / non-positive norm and `max_norm <= 0` are no-ops. The new test verifies both arms: cap-above-norm is untouched, small cap scales global norm to the cap.

**3. iterate_all_minibatches — PASS.** Uses the existing `generate_minibatch_indices_with_rng` helper (shuffle-once, chunk by `mb_size`) so there's no overlap/skip; the final remainder chunk is a valid smaller minibatch. Accumulated stats are divided by the actual `num_mb_steps`, so averaging is correct in both branches.

**4. New tests — PASS.** Both genuinely exercise the behavior (over-cap → scaled to cap, under-cap → unchanged; all-minibatch update produces finite stats) and pass.

**5. A/B evidence — sound.** Baseline reproduces the failure; the `BR_REWARD_SCALE=0.001` arm collapses BR value loss ~900 → ~8 while entropy stays pinned at the uniform-max floor (~1.228) and `avg_ap_loss` at `ln(40)`. That cleanly isolates and rules out H1 (gradient magnitude) as the binding constraint, pointing to critic-not-fitting / non-stationary opponent for the #239 follow-up. Methodology holds.

## Gates
- `cargo test --lib multi_agent --features training`: 69 passed, 0 failed, 3 ignored.
- Clippy clean on `training` and on `training,env-bucket-brigade` (examples).
- `cargo +nightly fmt --check`: clean.
- CI: all checks green (both Tests jobs Linux+macOS, Clippy, Rustfmt, Docs, Security Audit, Publish dry-run).
- Merge state: CLEAN / MERGEABLE.

Correct, well-tested, regression-safe infrastructure plus a decisive negative result. Lands the working grad-clip, all-minibatch iteration, and the VF_COEF/BR_TRAIN_STEPS knobs that the #239 follow-up will need.

Issue #241, opened 2026-06-24
Tooling: critic explained-variance logging + standalone single-best-response A/B harness for bucket-brigade


rjwalters:
## Motivation

Investigating why the PPO best-response (BR) doesn't learn on bucket-brigade (#239) is slow and blunt today: the only signal is the per-iteration `br[val= ent=]` line from a full NFSP run, and there is **no standalone single-BR harness** and **no critic explained-variance metric**. This issue builds the tooling that makes the #239 critic-fitting investigation tractable and tight.

## What to build

1. **Critic explained-variance logging.** In the BR PPO update (`src/multi_agent/joint.rs`), compute and log `explained_variance = 1 - Var(returns - values) / Var(returns)` per update (alongside the existing value-loss/entropy diagnostics surfaced by `examples/games/bucket_brigade/train_nfsp.rs:292-314`). EV near 0 (or negative) = critic not fitting; EV → 1 = critic fits. This is the single most informative missing metric for #239.
2. **Standalone single-best-response harness.** A small example (e.g. `examples/games/bucket_brigade/train_br_probe.rs`) that:
   - freezes N−1 opponents (random or a supplied fixed policy) for one cell,
   - trains ONE PPO best-response agent for K iterations,
   - logs per-iteration: value loss, **explained variance**, policy entropy, mean episode return.
   - is env-var configurable (CELL, ITERATIONS, ROLLOUT_STEPS, the #240 knobs VF_COEF / BR_TRAIN_STEPS / BR_REWARD_SCALE, plus a critic-LR knob if #239 adds one).
   This lets us A/B a single BR in ~1 min instead of via a full NFSP/PSRO run.

## Acceptance criteria

- [ ] Explained-variance is computed and logged in the BR update and surfaced in the bucket-brigade examples.
- [ ] A standalone single-BR harness exists, runs on one cell, and logs val-loss / EV / entropy / mean-return per iteration.
- [ ] Documented one-line command to run a BR A/B.
- [ ] No behavior change to existing trainers (additive logging + a new example); existing tests pass; clippy/fmt clean on both feature sets.

## Context

Part of #239 (PPO BR does not learn). The #240 infrastructure (wired grad-clip, all-minibatch iteration, VF_COEF/BR_TRAIN_STEPS knobs) is already merged. This is the enabler for the remaining #239 investigation.

PR #245, opened 2026-06-24
test(bucket-brigade): offline supervised regressor to diagnose BR critic fittability

### LINKED ISSUE #4: multi_agent::PolicyLearner has unresolved TODO placeholders

### LINKED ISSUE #242: Diagnose whether the bucket-brigade BR critic is structurally unfittable (partial observability) — per-agent vs centralized critic
## The pivotal question

The #239 investigation ruled out gradient magnitude as the cause of the BR not learning: even with `BR_REWARD_SCALE=0.001` (value loss collapsed ~900 → ~8), the value loss stays **flat** and entropy stays pinned at the uniform floor. The critic isn't fitting *even small, well-scaled targets*. The leading explanation is **partial observability**: a single agent's local observation may not contain enough information to predict its own/team return in the 4-player bucket-brigade game. If so, a per-agent PPO critic is **structurally unable** to fit, no amount of LR/steps tuning will help, and the answer is a **centralized critic** (CTDE: critic sees the joint state, actor stays decentralized).

This issue is the decision point: per-agent critic (fixable by tuning) vs centralized critic (architectural change).

## Diagnostic to run (cheap, offline)

1. Collect a rollout dataset on one cell (e.g. beta05) with frozen/random opponents: `(local_obs_i, discounted_return_i)` pairs.
2. **Fit a supervised regressor** (even a small MLP, many epochs, full-batch — no RL, no moving target) from `local_obs_i → return_i`. Measure explained variance on held-out data.
   - If the supervised critic ALSO can't fit (EV stays ~0): the return is **not predictable from the local observation** → partial observability confirmed → need a centralized critic.
   - If the supervised critic fits well (EV → 1): the target IS predictable; the RL critic's failure is a training-dynamics issue (LR/steps/non-stationarity) → pursue the per-agent fixes (#239 fix #4) instead.
3. Repeat feeding the **joint/global state** to the regressor; if that fits but local doesn't, that's direct evidence for a centralized critic.

## If centralized critic is indicated

Scope a follow-up: add a centralized value function (critic conditioned on joint observation / all agents' obs) used for advantage estimation while keeping decentralized actors (standard CTDE / MAPPO-style). This is a larger change; file it once the diagnostic confirms the need.

## Acceptance criteria

- [ ] Documented supervised-regression diagnostic on ≥1 cell: explained variance of `local_obs → return` vs `joint_state → return`.
- [ ] A clear conclusion: partial observability (→ centralized critic) vs training-dynamics (→ per-agent tuning), with the evidence.
- [ ] If centralized critic indicated: a follow-up issue filed with the design.

## Context

Part of #239. Best run after the explained-variance logging + single-BR harness (see the sibling tooling issue) land, which make this diagnostic much faster. Pivotal because the outcome decides the entire BR-fix strategy.

rjwalters:
## Summary

Resolves the #239/#242 decision point: is the per-agent best-response (BR) PPO **critic structurally unfittable** because of partial observability (→ a centralized CTDE critic is needed), or is its flat value loss a **training-dynamics** artifact (→ per-agent tuning)?

This PR adds a cheap, **offline** supervised-regression diagnostic (no RL loop, no moving target) and reports the measured explained-variance (EV) numbers. **The evidence points to training-dynamics, not partial observability — a centralized critic is NOT indicated for this cell.**

## What the diagnostic does

On the β=0.5 `minimal_specialization` cell (4 agents) with **uniform-random opponents** (the stationary "frozen opponents" baseline):

1. Collects a rollout dataset of `(local_obs_i, joint_actions, discounted_return_i)` for the focal agent. Returns are computed **offline** (`G_t = r_t + γ G_{t+1}`, no bootstrap / no value network) — a clean supervised target.
2. Fits a supervised MLP value-head regressor (the repo's proven `MlpBurnPolicy` value head — same architecture family as the failing RL critic) full-batch with Adam for 4000 epochs, on an **episode-shuffled** train/holdout split, tracking the **best held-out EV** (early-stopping proxy).
3. Repeats with the **joint state** (`local_obs ++ all agents' actions this step`) — a faithful CTDE-critic input proxy. If joint fits but local doesn't, that is direct evidence for a centralized critic.

It reuses the existing bucket-brigade `JointEnv` adapter + flat-obs encoding for rollout collection (mirrors `make_cell_env` / the rollout loop in `tests/test_psro_bucket_brigade.rs`), but the regressor and EV computation are local so it is **self-contained** — it does **not** depend on the unmerged tooling issue #241.

## Measured results (β=0.5, 40 episodes × 200 steps, random opponents)

| Input | train MSE (norm) | **best holdout EV** | final holdout EV |
|-------|------------------|---------------------|------------------|
| `local_obs → return` (47-d) | 1.078 → 0.009 | **0.491** | -0.341 |
| `joint(obs+actions) → return` (59-d) | 1.020 → 0.001 | **0.463** | -0.559 |

- **Joint improvement over local: Δ ≈ -0.03** (the joint state adds nothing).
- Cross-checked at 80 episodes (10,839 samples): local best holdout EV ≈ 0.43 — same picture.

Dataset stats: 5,781 samples, return mean ≈ -12,589, std ≈ 6,138 (returns are large because γ=0.99 over a 200-night horizon at ≈-63 team reward/night).

## Conclusion (acceptance criterion 2)

**Training-dynamics, NOT partial observability. A centralized critic is not indicated.**

- The local observation already explains **~half** the held-out return variance (EV ≈ 0.49). It is *not* the case that a single agent's local obs carries no information about its return — the partial-observability hypothesis is **rejected** for this cell.
- Adding the other agents' actions (the CTDE information a centralized critic would supply) **does not help** (Δ ≈ -0.03), so the missing-information story does not hold.
- The regressor overfits hard (train EV → ~0.99 while held-out EV peaks at ~0.49 then *degrades* to negative by the final epoch). The per-agent critic's headroom is therefore bounded by **overfitting / non-stationarity / target scale**, not by a structural information bottleneck. → pursue the per-agent fixes (#239 fix #4: LR/steps/target-scaling/regularization), not a CTDE rearchitecture.

Because the diagnostic does **not** indicate a centralized critic, **no centralized-critic follow-up issue is filed** (acceptance criterion 3 is conditional on the centralized-critic outcome).

### Methodology note worth recording

An episode-**contiguous** holdout split gave strongly negative EV for both inputs and would have falsely read as "unpredictable". The cause is that per-episode return-level offsets (random opponents make each episode's overall fire severity vary wildly: return std ≈ 6000 on a mean of ≈ -12,600) dominate the variance and don't transfer across held-out episodes. **Shuffling** the split isolates the within-state obs→return signal — which is exactly what the critic needs for advantage estimation. The committed test shuffles and reports best-EV for this reason.

### Burn gotcha found & worked around

A hand-rolled `#[derive(Module)]` regressor's params were silently **not** collected by `GradientsParams::from_module`/`from_grads` (the optimizer stepped nothing — training loss frozen to 6 decimals, even on a trivially-learnable synthetic target). Reusing the repo's proven `MlpBurnPolicy` value head fixed it. Documented inline so the next person doesn't lose time on it.

## Changes

- Add `tests/test_bucket_brigade_critic_fittability.rs`: an `#[ignore]`-gated integration test (`diagnose_br_critic_fittability_beta05`) that runs the full diagnostic and asserts the decision-rule invariants (local EV must be clearly positive; joint must not generalize dramatically worse than local). Run with:
  ```bash
  cargo test --features "training,env-bucket-brigade" --release \
      --test test_bucket_brigade_critic_fittability -- --ignored --nocapture
  ```

| Documented supervised-regression diagnostic on ≥1 cell: EV of `local_obs→return` vs `joint_state→return` | ✅ | β=0.5 cell; local best EV 0.491, joint best EV 0.463 (table above; logged by the test) |
| A clear conclusion (partial-obs vs training-dynamics) WITH the evidence | ✅ | Training-dynamics; evidence = local EV ≈ 0.49 (not ~0) and joint Δ ≈ -0.03 (no CTDE gain) |
| If centralized critic indicated: file follow-up | ✅ (n/a) | Centralized critic is NOT indicated, so no follow-up filed (criterion is conditional) |

## Test Plan

- `cargo fmt -- --check` — clean.
- `cargo clippy --all-targets --features "training,env-bucket-brigade" -- -D warnings` — clean.
- `cargo test --features "training,env-bucket-brigade" --release --test test_bucket_brigade_critic_fittability -- --ignored` — **1 passed** (assertions encode the decision rule).

Closes #242

### DIFF
--- a/tests/test_bucket_brigade_critic_fittability.rs
+++ b/tests/test_bucket_brigade_critic_fittability.rs
@@ -1,0 +1,558 @@
+//! Offline supervised-regression diagnostic for the bucket-brigade
+//! best-response (BR) critic (issue #242, part of #239).
+//!
+//! # The question
+//!
+//! The #239 investigation showed the per-agent PPO BR **critic value loss
+//! stays flat** even with tiny, well-scaled targets (`BR_REWARD_SCALE=0.001`,
+//! value loss collapsed ~900 → ~8 but never *fit*). Two competing
+//! explanations remain:
+//!
+//! 1. **Partial observability** — a single agent's *local* observation does not
+//!    contain enough information to predict its own discounted return in the
+//!    4-player game. If so, a per-agent critic is *structurally* unable to fit
+//!    and no LR/steps tuning will help → the answer is a **centralized critic**
+//!    (CTDE: critic sees the joint state, actor stays decentralized).
+//! 2. **Training dynamics** — the target *is* predictable from local obs, and
+//!    the RL critic's failure is an optimization/non-stationarity artifact →
+//!    pursue per-agent fixes (LR/steps/target-scaling).
+//!
+//! This is the **decision point** between those two strategies.
+//!
+//! # The diagnostic (cheap, offline — NO RL loop, NO moving target)
+//!
+//! 1. Collect a rollout dataset on one cell (`beta05`) with **uniform-random
+//!    opponents** (the "frozen opponents" baseline): for the focal agent we
+//!    record `(local_obs_i, joint_state, discounted_return_i)` triples, where
+//!    `discounted_return_i = Σ_{t'>=t} γ^{t'-t} r_{i,t'}` is computed *offline*
+//!    from the realized reward stream (no value bootstrap, no moving target).
+//! 2. Fit a small supervised MLP regressor `local_obs_i → return_i` with
+//!    full-batch Adam for many epochs (a clean supervised problem — the only
+//!    thing that can stop it fitting is the *information content* of the
+//!    input). Measure **explained variance (EV)** on a held-out split.
+//! 3. Repeat with the **joint state** as input.
+//!
+//! Decision rule:
+//!
+//! - local EV ~0 and joint EV ≫ local EV → return is **not predictable from
+//!   local obs** but **is** from the joint state → partial observability
+//!   confirmed → **centralized critic needed**.
+//! - local EV substantially positive and joint adds little → the target *is*
+//!   predictable from local obs → the RL critic failure is **training
+//!   dynamics**, not missing information → pursue per-agent tuning.
+//!
+//! # Measured result (β=0.5, 40 episodes × 200 steps, random opponents)
+//!
+//! With an episode-**shuffled** train/holdout split and early-stopping (best
+//! held-out EV across epochs):
+//!
+//! - `local_obs → return`: **best held-out EV ≈ 0.48**
+//! - `joint(obs+actions) → return`: **best held-out EV ≈ 0.46** (Δ ≈ −0.02)
+//!
+//! The local observation already explains ~half the held-out return variance,
+//! and adding the other agents' actions does **not** improve it. This is the
+//! **opposite** of the partial-observability hypothesis: the return is *not*
+//! hidden from the local critic. The RL BR critic's flat value loss (#239) is
+//! therefore a **training-dynamics** problem — note that the supervised
+//! regressor itself overfits hard (train EV → ~0.99 while held-out EV peaks at
+//! ~0.48 then *degrades* to negative by the final epoch), so the per-agent
+//! critic's headroom is bounded by overfitting/non-stationarity, not by a
+//! structural information bottleneck. A centralized critic is **not** the
+//! indicated fix for this cell.
+//!
+//! Caveat worth recording: the original episode-**contiguous** split gave
+//! strongly negative EV for both inputs, because per-episode return-level
+//! offsets (random opponents make each episode's overall fire severity vary
+//! wildly: return std ≈ 6000 on a mean of ≈ −12600) dominate the variance and
+//! do not transfer across held-out episodes. Shuffling isolates the
+//! within-state signal, which is what the critic needs for advantages.
+//!
+//! # Self-contained by design (issue #242 constraint)
+//!
+//! The sibling tooling issue #241 (explained-variance logging + single-BR
+//! harness) is **not merged**. This diagnostic deliberately does **not**
+//! depend on it: it reuses the existing bucket-brigade `JointEnv` adapter and
+//! flat-observation encoding for rollout collection, but the regressor and
+//! the EV computation are implemented locally here so the experiment runs
+//! standalone.
+//!
+//! # The "joint state" input
+//!
+//! The bucket-brigade per-agent flat observation is *almost* fully global
+//! already (houses / signals / locations / last-actions / scenario-info are
+//! shared state; only the leading `agent_id_norm` scalar differs per agent —
+//! see the `flatten_observation` docstring in
+//! `src/env/games/bucket_brigade/env.rs`). The genuinely *private* extra
+//! information a centralized critic would add is **the other agents' chosen
+//! actions this step**. So we build the "joint state" input as
+//! `local_obs ++ all_agents_actions_this_step`, which is a faithful proxy for
+//! the CTDE critic's input (joint observation + joint action). If that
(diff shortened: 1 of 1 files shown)


rjwalters:
✅ **Approved!**

Thorough evaluation of this diagnostic/research PR against the #242 acceptance criteria. The conclusion drives a major BR-fix strategy decision, so I scrutinized the methodology and reproduced the numbers independently.

## Acceptance criteria
| Criterion | Status |
|-----------|--------|
| Supervised-regression diagnostic on ≥1 cell: EV(local_obs→return) vs EV(joint→return) | ✅ β=0.5 cell; reproduced locally |
| Clear evidence-backed conclusion (partial-obs vs training-dynamics) | ✅ Training-dynamics; evidence supports it |
| Follow-up issue (conditional on centralized-critic outcome) | ✅ n/a — training-dynamics concluded, correctly no follow-up |

## Verification performed
- **Reproduced the test** (`--ignored --nocapture`): local best holdout EV = **0.478**, joint = **0.453**, Δ = **-0.025**. 1 passed in ~90s. Matches the PR table.
- **EV formula is correct**: `1 - ss_res/ss_tot`, with predictions de-normalized to original target units before computing residuals against `test_y`. Train/holdout split uses train-only standardization stats (no leakage).
- **Training is real, not a frozen-params artifact**: train MSE(norm) drops 1.0 → 0.008. The `MlpBurnPolicy` value-head reuse is the right workaround for the documented Burn `GradientsParams::from_module` gotcha — confirmed `forward(x).1` (value head, shape `[batch]`) is read and the model is genuinely Adam-stepped.
- **Self-contained**: reuses only `JointEnv` + `flatten_observation`; no dependency on unmerged #241. Confirmed.
- **Joint-state proxy is faithful**: env docstring confirms the flat obs is almost-global (only `agent_id_norm` differs per agent), so `local_obs ++ joint_actions` is a sound CTDE input — the "joint adds nothing" result is well-grounded.

## Pivotal scientific check (the shuffle-vs-contiguous question)
I independently reran with the **contiguous** split (shuffle disabled) to test whether shuffling masks a partial-observability signal. It does **not**:
- contiguous local best EV ≈ **0.23** (not ~0), joint ≈ **0.30**, gain ≈ **+0.066** (well under the 0.25 threshold).
- The partial-observability hypothesis (local≈0, joint≫local) is **rejected under both splits**. The conclusion does not hinge on the shuffle choice — it is robust.

Shuffling is scientifically defensible: GAE/PPO advantage estimation needs the *within-state* obs→return signal, while the per-episode return-level offset (exogenous fire stochasticity, std≈6000 on mean≈-12,600) is a constant the baseline subtracts. Isolating the within-state signal is the correct target.

## Minor (non-blocking) — tracked in #246
The note says the contiguous split gave "strongly negative EV for both inputs." That's true of *final-epoch* EV, but the harness tracks *best* EV, which stays mildly positive (~0.23/~0.30) even contiguous. Documentation-accuracy nit only; filed #246. Conclusion unaffected.

## CI / merge
- All checks pass (Clippy, Rustfmt, Docs, both Tests jobs, Security Audit, Publish dry-run).
- `cargo fmt --check` and `cargo clippy -D warnings` clean locally.
- Merge state: CLEAN. `Closes #242` present.

No scope creep — single `#[ignore]`-gated test file, 558 additions, 0 deletions. Solid, well-documented research deliverable.

PR #298, opened 2026-07-06
feat: recurrent PPO trainer + FlickeringCartPole POMDP with LSTM-vs-MLP memory contrast

rjwalters:
Closes #287

Phase 3 of the recurrent-policy epic (#262), implemented per the orchestrator re-scope on the issue: the learning-signal environment is **FlickeringCartPole** (whole-observation dropout at p=0.5, Hausknecht & Stone 2015 protocol), with **MaskedCartPole retained as a documented negative result**.

## What's in this PR

**Commit 629f01c (first attempt, validated stack):**
- `src/train/ppo/recurrent_trainer.rs` — `RecurrentPPOTrainer<B,P,O>`: rank-3 sibling of `PPOTrainerBurn`; consumes `RecurrentRolloutBuffer` sequence batches, flattens `[N_env,T]` -> rank-1 for the unchanged loss fns, consumes `episode_starts` directly, caller-driven LR annealing. Unit-tested on a synthetic rank-3 batch.
- `src/env/games/masked_cartpole.rs` + example — retained as the empirical record that velocity-masking is NOT a POMDP (memoryless MLP 324.5 beat LSTM 221.9 over 500k steps; see issue thread). Doc comments state the negative result explicitly.

**Commit cfa2c35 (this re-scope):**
- `src/env/games/flickering_cartpole.rs` — composition over `CartPole`; with seeded probability `p` (default 0.5) the entire 4-D observation is zeroed. Physics/reward/termination unchanged. Snapshot/restore captures the flicker RNG. 12 unit tests: determinism under seed, blank rate ~ p, shape invariance, reward/termination unaffected, clone/restore reproduces the flicker stream.
- `examples/games/cartpole/recurrent_ppo_flickering_cartpole.rs` — end-to-end LSTM-vs-MLP contrast on the identical flickering stream (same seeds, same normalizer, same budget; only the policy class differs).

## Measured results (500k env steps per arm, p=0.5, seed 0, CPU release)

| Policy | Final mean return (last <=100 eps) | Peak | Bar |
|---|---|---|---|
| **LSTM (RecurrentPPOTrainer)** | **467.6** | **483.9** | >195: **CLEARED** (crosses ~195 by ~50k steps) |
| MLP (feedforward, same stream) | 163.0 | 177.4 | stays below bar and far below LSTM |

[...]

PR #300, opened 2026-07-06
fix(ppo): honor PPOConfig::max_grad_norm in both PPO trainers

### LINKED ISSUE #299: PPO trainers silently ignore PPOConfig::max_grad_norm
Found during #287 (PR #298): both PPO trainers read `max_grad_norm` from `PPOConfig` but never apply it — only SAC and A2C apply gradient clipping. This caused a real failure: on FlickeringCartPole, disabling value clipping blew up the shared LSTM trunk because nothing bounded the gradients. The #298 example works around it at the optimizer level with `GradientClipping::Norm(0.5)`.

Either honor the config field in both PPO trainers (recommended — SAC/A2C show the pattern) or remove the field and document clipping as example-level responsibility. Silent no-op config is the worst of the options.

Evidence: see the run 1–3 contrast table on #287 (runs 1–2 vs run 3).

rjwalters:
Closes #299

## What

Both PPO trainers (`PPOTrainerBurn`, `RecurrentPPOTrainer`) read `PPOConfig::max_grad_norm`, validated it, and then silently ignored it — only SAC/A2C/joint applied any gradient clipping. This PR honors the config field in both trainers, matching the joint multi-agent trainer's pattern (issue #239).

- **`src/train/grad_clip.rs` (new)**: extracted `clip_grads_by_global_norm` (and a `global_grad_norm` helper) verbatim from `src/multi_agent/joint.rs`, where it was a private fn, so all three call sites share one definition. It clips the **global** L2 norm of the concatenated per-parameter gradients (SB3/PyTorch `clip_grad_norm_` semantics), preserving gradient direction — unlike Burn's built-in `GradientClipping::Norm`, which clips each parameter tensor independently.
- **`PPOTrainerBurn::new` / `RecurrentPPOTrainer::new`**: stage `config.max_grad_norm` on the `BurnOptimizer` wrapper (same as A2C/joint); `train_step` applies the clip to the gradient slice immediately before every move-through `inner_mut().step(...)`.
- **`src/multi_agent/joint.rs`**: now imports the shared helper; its unit test moved to `train::grad_clip::tests` unchanged.
- **`PPOConfig::max_grad_norm` doc comment** now states exactly where and how the clip is applied (criterion 4).
- **Example migration** (criterion 5): `recurrent_ppo_flickering_cartpole` dropped its optimizer-level `GradientClippingConfig::Norm(0.5)` workaround (both arms) and now relies on the `.max_grad_norm(0.5)` it already set in `PPOConfig`. Compile-checked (`cargo clippy --all-targets`); not retrained.

## Default-behavior decision (criterion 2 / regression note)

`max_grad_norm` was already a non-optional `f64` with default `0.5`, and `validate()` rejects non-positive values — i.e. the field's existing default (and every existing call site, all of which set `0.5`) **already promised clipping**. Per the curation note, this PR therefore keeps the existing type and default and documents the change as a **behavior fix**, not a silent default change:

- Configs that left the default or set `.max_grad_norm(0.5)` now actually get the clipping they were asking for. This can change training trajectories — that is the bug fix (PR #298 showed the absence of clipping blows up shared LSTM trunks).
- There is no `None`/"disable" state to preserve because the type never had one; users who genuinely want no clipping can set a very large cap (e.g. `1e9`), which is documented on the field.
- Clip semantics note: the trainers apply a *global*-norm clip (direction-preserving), while the #298 example workaround used Burn's *per-tensor* norm clip. The global-norm variant is the standard PPO clip and the one the joint trainer already uses; the example's behavior is therefore not bit-identical to the workaround, but is the intended semantics of the config field.

## Tests (criterion 3)

- `train::grad_clip::tests::test_clip_grads_by_global_norm` (moved from joint.rs): cap above the gradient norm is a no-op (no-clip control); cap below scales the global norm to the cap.
- `ppo_trainer_burn_applies_max_grad_norm` / `recurrent_trainer_applies_max_grad_norm` (new): two trainers from cloned policies, identical data/seeds, differing only in the cap — a tiny cap (`1e-6`) must produce a parameter update < 0.2x the effectively-unbounded control (`1e9`).

## Verification

- `cargo test` — full suite green
- `cargo clippy --all-targets -- -D warnings` — clean (also compile-checks all examples, including the migrated one)
- `cargo fmt --check` — clean
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps` — clean

### DIFF
--- a/examples/games/cartpole/recurrent_ppo_flickering_cartpole.rs
+++ b/examples/games/cartpole/recurrent_ppo_flickering_cartpole.rs
@@ -47,7 +47,6 @@ use std::sync::atomic::{AtomicU64, Ordering};
 use anyhow::Result;
 use burn::{
     backend::Autodiff,
-    grad_clipping::GradientClippingConfig,
     nn::LstmState,
     optim::AdamConfig,
     tensor::{Int, Tensor, TensorData, activation},
@@ -239,13 +238,11 @@ fn run_lstm(total_timesteps: usize, flicker_prob: f64) -> Result<ArmResult> {
         LstmBurnPolicy::<Backend>::with_config(obs_dim, action_dim, policy_config, &device);
 
     let lr = LSTM_LR;
-    // Global gradient-norm clip at the optimizer level: the PPO trainers do
-    // not themselves apply `max_grad_norm`, so honor it here (matches SAC's
-    // `build_adam`). Load-bearing for the recurrent trunk, where value-loss
+    // Gradient clipping comes from `PPOConfig::max_grad_norm` below (issue
+    // #299): the trainer applies a global gradient-norm clip before every
+    // optimizer step. Load-bearing for the recurrent trunk, where value-loss
     // spikes otherwise wipe out the policy features.
-    let inner_opt = AdamConfig::new()
-        .with_grad_clipping(Some(GradientClippingConfig::Norm(0.5)))
-        .init();
+    let inner_opt = AdamConfig::new().init();
     let burn_opt: BurnOptimizer<Backend, LstmBurnPolicy<Backend>, _> =
         BurnOptimizer::new(inner_opt, lr);
 
@@ -474,10 +471,9 @@ fn run_feedforward(total_timesteps: usize, flicker_prob: f64) -> Result<ArmResul
     };
     let policy = MlpBurnPolicy::<Backend>::with_config(obs_dim, action_dim, policy_config, &device);
 
-    // Gradient clipping mirrors the LSTM arm (see run_lstm).
-    let inner_opt = AdamConfig::new()
-        .with_grad_clipping(Some(GradientClippingConfig::Norm(0.5)))
-        .init();
+    // Gradient clipping comes from `PPOConfig::max_grad_norm` below,
+    // mirroring the LSTM arm (see run_lstm).
+    let inner_opt = AdamConfig::new().init();
     let burn_opt: BurnOptimizer<Backend, MlpBurnPolicy<Backend>, _> =
         BurnOptimizer::new(inner_opt, MLP_LR);
 
--- a/src/multi_agent/joint.rs
+++ b/src/multi_agent/joint.rs
@@ -68,13 +68,14 @@
 
 use anyhow::{Result, anyhow};
 use burn::{
-    module::{AutodiffModule, Module, ModuleVisitor, Param},
+    module::AutodiffModule,
     optim::{GradientsParams, Optimizer},
     tensor::{Int, Tensor, backend::AutodiffBackend},
 };
 use rand::rngs::StdRng;
 
 use crate::train::{
+    grad_clip::clip_grads_by_global_norm,
     optimizer::{BackendOptimizer, BurnOptimizer},
     ppo::loss::{
         compute_policy_loss, compute_value_loss, generate_minibatch_indices_with_rng, scalar_f64,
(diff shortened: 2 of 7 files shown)


rjwalters:
✅ **Approved!**

Thorough review of the `max_grad_norm` fix, including the two judgment calls flagged in the PR body.

## Judgment call 1: default-behavior change — correct call

The builder invoked the curation note's escape clause ("unless the field's existing default already promises clipping") and I confirm it applies:

- `PPOConfig::max_grad_norm` is a non-optional `f64` defaulting to `0.5`, and `validate()` rejects non-positive values — the type has **no representable "no clip" state**. The field's contract has always promised clipping at 0.5; the trainers were simply not honoring it.
- I verified every PPO call site in the repo (`train_cartpole_modern`, `train_cartpole_async`, `train_pong_self_play`, `train_snake_multi_v2`, both recurrent examples, `tests/test_cartpole_async.rs`) explicitly sets `.max_grad_norm(0.5)`. None sets a large value to opt out; nothing relies on effectively-unclipped gradients. (The `10.0` values are `DQNConfig`, out of scope.)
- Blast-radius spot-check: I ran the learning-sensitive `cartpole_async_reaches_learning_bar` integration test locally with clipping now active — **passes** (return ≥ 60 bar cleared, 14.6s). `test_bc_cartpole` smoke and all 15 `multi_agent::joint` lib tests also pass locally.
- The `1e9` disable escape hatch is documented on the field and is accurate given `validate()`'s positivity requirement.

Documenting this as a behavior fix (not a silent default change) is the right framing.

## Judgment call 2: clip-semantics swap in the example — acceptable, tracked in #301

The example's removed workaround was Burn's **per-tensor** `GradientClipping::Norm(0.5)`; the trainer now applies a **global**-norm clip at 0.5. These are not numerically identical, but the swap is safe in the direction that matters: a global clip at cap `c` is never *looser* than a per-tensor clip at `c` (per-tensor permits total step norm up to `sqrt(K)·c`; global caps it at `c`). The load-bearing property — bounding value-loss spikes through the shared LSTM trunk (PR #298) — is preserved or strengthened, and direction preservation is exactly what the example's own comment says it needs. The residual risk is strictly-tighter clipping slowing learning below the 195 bar.

Shipping without a 2×500k-step CPU retrain is reasonable because (a) the example is compile-checked, not learning-gated, in CI, (b) the #298 result numbers live in the PR/issue record, not in shipped repo docs, and (c) the new semantics is the intended meaning of the config field. **Created #301** to retrain the example under the new semantics and refresh the reference numbers.

## Code review

- **Extraction from `joint.rs`**: verified line-by-line — pure code motion. Only deltas: pass 1 factored into the new `global_grad_norm` helper (logic identical, same `1e-6` eps, same finite/positivity guards), doc updated to mention #299, and the moved unit test now uses the shared norm helper instead of its local duplicate. No semantic drift. `joint.rs` call sites unchanged.
- **Trainer integration**: staging `config.max_grad_norm` on the `BurnOptimizer` wrapper in `new()` and reading `grad_clip_norm()` back before each `inner_mut().step(...)` exactly mirrors the A2C/joint pattern. Clip applied per minibatch step in both trainers, before the move-through step — correct placement. `RecurrentPPOTrainer`'s `lr_override` path preserved.
- **No double-clip risk**: repo-wide grep confirms no remaining PPO call site builds its Burn optimizer with `with_grad_clipping` (only SAC, which manages its own optimizers).
- **Tests**: the clipped-vs-control methodology is sound — cloned policies, identical data/seeds, exactly one gradient step (`n_epochs(1)`, single minibatch; `target_kl(1.0)` on the recurrent test to defeat early-stop), only the cap differs; `1e-6` cap vs `1e9` control with a < 0.2x update-norm assertion. Both pass locally, as do the moved `grad_clip` tests.
- **Docs**: `PPOConfig::max_grad_norm` doc comment accurately states where/how the clip is applied and the disable escape hatch (criterion 4). New `train::grad_clip` module and re-exports are documented.

## CI / merge state

- All 9 CI checks **pass** (including Tests Linux 11m47s, Tests macOS 8m32s full suites).
- `mergeStateStatus`: **CLEAN**.
- PR body uses `Closes #299`.

All six curation acceptance criteria met. Nice, disciplined fix.

PR #317, opened 2026-07-07
fix(psro): serialize best-response backward passes to dodge burn-autodiff deadlock

### LINKED ISSUE #307: Local deadlock: PSRO parallel best-response wedges in burn-autodiff GraphLocator/GraphState mutexes (test_psro_n_player_matching_pennies)
## Symptom

On a macOS arm64 host (Darwin 25.5.0, 8-core M-series), `cargo test --features training --test test_psro_n_player_matching_pennies` deterministically wedges at ~0% CPU. Reproduced 4/4 attempts, including:

- the full suite run (both smoke tests parked >25 min, cumulative CPU 2.1s),
- the binary in isolation,
- `--test-threads=1` (a single test, `psro_n4_issue215_knobs_run_and_are_finite`, running alone),
- with and without the Claude Code sandbox (so it is not a sandbox artifact).

CI on the same commit (`c5d8fd6`, Linux + macOS GitHub runners) is green, so this is timing/machine dependent, not a deterministic code regression.

## Evidence (thread sample of the wedged process)

`sample` of the single-test run shows all threads parked:

- main thread: `std::thread::park` (test harness waiting for the test thread),
- test thread `psro_n4_issue215_knobs_run_and_are_finite`: `__psynch_cvwait` inside the rayon parallel region of `PsroTrainer::train_best_responses_parallel` (src/multi_agent/psro.rs, issue #232 parallel BR loop),
- multiple rayon worker threads: blocked in `parking_lot::RawMutex::lock_slow` acquiring **`burn_autodiff::runtime::graph::GraphLocator`** and **`burn_autodiff::runtime::graph::GraphState`** mutexes,
- remaining rayon workers: `WorkerThread::wait_until_out_of_work` (idle).

i.e. a lost-wakeup/lock-order deadlock between the rayon-parallel best-response training (multiple concurrent `Autodiff<NdArray>` backward graphs) and burn-autodiff's internal graph locking (burn 0.21).

## Why filed

Found while running the full suite for #304 (whose diff is an example + docs only — it cannot interact with this code path). Builder protocol: pre-existing/unrelated failures get their own issue rather than in-PR fixes.

## Possible directions

- Reproduce under stress on other hosts (loop the smoke test N times; CI runners may simply be winning the race).
- Check burn 0.21 upstream for known autodiff graph-lock deadlocks under concurrent backwards on distinct graphs; consider pinning/bumping.
- Consider serializing the per-agent BR `update` calls (rollouts stay parallel) or gating the parallel BR path behind a config knob when running on the global pool.

rjwalters:
## Summary

Fixes the local N=4 PSRO deadlock (#307) where `test_psro_n_player_matching_pennies` deterministically wedges at ~0% CPU on macOS arm64. The root cause is a lock-order inversion between the `#232` rayon-parallel best-response training (four concurrent `Autodiff<NdArray>` backward passes) and `burn-autodiff 0.21`'s internal `GraphLocator`/`GraphState` `parking_lot` mutexes.

**Approach chosen: config knob (Direction C1), not a burn bump (Direction B).** Per the curator's guidance I checked crates.io first — `burn`/`burn-ndarray` max out at `0.21.0` (verified via `cargo info burn` and the crates.io API: `max_version: 0.21.0`). No `0.22` exists, so the upstream bump that would fix the deadlock at the source is not yet possible. The knob is the fully-reversible immediate fix; set it back to `false` once a fixed burn is adopted.

## Changes

- Add `PsroConfig::serialize_br_updates: bool` (default `true`), documented with the #307 root cause, the determinism rationale, and a note to restore `false` after the burn bump.
- Branch `PsroTrainer::train_best_responses_parallel`: bind the per-agent BR closure once and dispatch via serial `.map()` when `serialize_br_updates` is set, else the original `into_par_iter().map()`. Only backward-pass concurrency changes; the fixed-order RNG pre-pass and collect-by-index join are untouched. Not gated behind `cfg(target_os)`.
- Add regression guard `psro_n4_serialized_br_is_deterministic_issue307`: runs two seeded N=4 runs and asserts bit-identical NashConv + per-agent meta-Nash marginals (guards both hang-free completion and the #232 determinism invariant on every CI run).
- Propagate the new field (`true`) to every explicit `PsroConfig { .. }` construction across `src/`, `tests/`, and the bucket-brigade example.

| `test_psro_n_player_matching_pennies` completes < 3 min, 5 consecutive runs, no `__psynch_cvwait` stall (macOS arm64) | ✅ | 5-run loop of the fast test: all `ok`, 17–24s each; full binary also completes in ~30–88s |
| Same command passes with `--test-threads=1` | ✅ | Ran `-- --test-threads=1`: 2 passed, finished in 31.31s |
| `cargo test --features training` stays green on Linux CI | ✅ | Full suite green locally (500 lib + all integration binaries, 0 failed); change is host-agnostic |
| `psro_n4_smoke_runs_and_is_finite` + `psro_n4_issue215_knobs_run_and_are_finite` pass in default config | ✅ | Both pass under default (`serialize_br_updates: true`) |
| Result unchanged for a fixed seed (determinism / #232 preserved) | ✅ | New `psro_n4_serialized_br_is_deterministic_issue307` asserts bit-identical `PsroStats` across two seed=19 runs; passes |
| `serialize_br_updates` documented with #307 ref + when to reset to `false` | ✅ | Doc comment on the field |
| burn version bump | N/A | No burn 0.22 on crates.io; Direction C1 used instead |

## Test Plan

- `cargo test --features training` — full suite green (all binaries 0 failed).
- `test_psro_n_player_matching_pennies` run 5 consecutive times + once with `--test-threads=1` — no hang, all pass.
- `cargo clippy --all-targets --features training -- -D warnings` — clean.
- `cargo fmt --check` — clean.
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --features training` — clean.

Closes #307

### DIFF
--- a/examples/games/bucket_brigade/train_psro.rs
+++ b/examples/games/bucket_brigade/train_psro.rs
@@ -332,6 +332,9 @@ fn main() -> Result<()> {
         max_payoff_evals_per_iteration,
         br_reward_scale,
         seed: base_seed(),
+        // Serialize per-agent BR backward passes to avoid the burn-autodiff
+        // 0.21 graph-lock deadlock on many-core hosts (issue #307).
+        serialize_br_updates: true,
     };
     let joint_config = JointTrainerConfig {
         num_agents: NUM_AGENTS,
--- a/src/multi_agent/psro.rs
+++ b/src/multi_agent/psro.rs
@@ -1002,6 +1002,40 @@ pub struct PsroConfig {
     pub br_reward_scale: f32,
     /// RNG seed for opponent sampling and deterministic tests.
     pub seed: u64,
+    /// Serialize the per-agent best-response `update` (backward) calls to
+    /// avoid a concurrent-`backward()` deadlock in `burn-autodiff 0.21` on
+    /// some many-core hosts (issue #307).
+    ///
+    /// The #232 parallel path dispatches `num_agents` independent
+    /// best-response tasks via `rayon`, so all agents' `joint_loss.backward()`
+    /// passes run concurrently on worker threads. `burn-autodiff 0.21`
+    /// guards its `GraphLocator`/`GraphState` singletons with `parking_lot`
+    /// mutexes acquired in an inconsistent order across the two backward
+    /// code paths, producing a lock-order inversion that deterministically
+    /// wedges on macOS arm64 (8-core M-series) with `N = 4` concurrent
+    /// backward passes. Linux CI (2-core x86) rarely hits the race window.
+    ///
+    /// When `true` (the default until the upstream burn issue is resolved
+    /// — no `burn 0.22` exists on crates.io yet), the parallel `par_iter`
+    /// is replaced by a serial `.map()` that trains each agent's BR in
+    /// fixed agent order, so at most one backward graph is live at a time.
+    /// When `false`, all agents' backward passes run concurrently under
+    /// `rayon` (the original #232 path).
+    ///
+    /// # Determinism
+    ///
+    /// This flag does **not** affect results. All shared-mutable draws are
+    /// hoisted into the fixed-order pre-pass in
+    /// `train_best_responses_parallel` before any dispatch, and each
+    /// BR is a pure function of its pre-drawn inputs, collected by index.
+    /// Serial vs. parallel dispatch therefore yields bit-identical
+    /// `PsroStats` for a given seed; only wall-clock concurrency of the
+    /// backward passes changes. The #232 thread-count-invariance guarantee
+    /// is preserved in both modes.
+    ///
+    /// Set this back to `false` once a burn version that fixes the
+    /// autodiff graph-lock inversion is adopted (tracked in #307).
+    pub serialize_br_updates: bool,
 }
 
 impl Default for PsroConfig {
@@ -1014,6 +1048,7 @@ impl Default for PsroConfig {
             max_payoff_evals_per_iteration: None,
             br_reward_scale: 1.0,
             seed: 0,
+            serialize_br_updates: true,
         }
     }
 }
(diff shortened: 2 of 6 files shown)


rjwalters:
✅ **Approved!**

Thorough, well-targeted fix for the #307 burn-autodiff deadlock. The config-knob approach (Direction C1) is the right call given no `burn 0.22` exists on crates.io yet, and the implementation is clean and fully reversible.

## Correctness

- The BR closure is bound once (`run_br`) and dispatched via serial `.map()` when `serialize_br_updates` is set, else the original `into_par_iter().map()`. Only backward-pass concurrency changes — the fixed-order RNG pre-pass and collect-by-index join are untouched, so the serialized path holds at most one `Autodiff` backward graph live at a time, side-stepping the `GraphLocator`/`GraphState` lock-order inversion.
- No `cfg(target_os)` gating — the fix is host-agnostic, as required by the acceptance criteria. The serialized path is safe everywhere; Linux just rarely hit the race window.

## Default flip is safe for all constructions

`PsroConfig` is a plain `pub struct` (no `#[non_exhaustive]`, no `..Default::default()` shortcuts), so every explicit literal must carry the new field. Verified all 12 explicit constructions + the `Default` impl now set `serialize_br_updates: true`:
- `src/multi_agent/psro.rs`: Default impl + 4 test literals
- `tests/`: test_psro_n_player_matching_pennies, test_psro_bucket_brigade, test_psro_matching_pennies (x3), test_seeded_reproducibility
- `examples/games/bucket_brigade/train_psro.rs`

Green CI confirms nothing was missed at compile time.

## Determinism (#232 preserved)

The new `psro_n4_serialized_br_is_deterministic_issue307` regression guard runs two seeded N=4 runs and asserts bit-identical NashConv (`to_bits()`) and per-agent meta-Nash marginals. This pins both hang-free completion and the #232 thread-count-invariance guarantee on every CI run — a strong, permanent guard rather than a one-off manual check.

## Documentation

The field doc comment is excellent: it explains the burn-0.21 root cause (concurrent `backward()` on distinct graphs, `parking_lot` mutex lock-order inversion), the determinism rationale, and the explicit revert path ("set back to `false` once a burn version that fixes the graph-lock inversion is adopted, tracked in #307").

## CI Status

All 9 checks pass, including the load-bearing ones:
- Tests (Linux + Burn NdArray) — pass (13m1s)
- Tests (macOS + Burn NdArray) — pass (13m55s)
- Clippy, Rustfmt, Documentation, Security Audit, Publish dry-run — all pass

Merge state: CLEAN.

## Test Execution

Relied on GitHub CI (both macOS and Linux full `--features training` suites) rather than re-running the 13-14 min suites locally, since CI already exercises the exact deadlock-prone path on both platforms and is green.

## Minor note (non-blocking)

Adding a public field to `PsroConfig` is technically a semver-breaking change for any downstream crate constructing it with a struct literal. That's fine pre-1.0 (crate is at 0.3.0). A future `#[non_exhaustive]` on config structs would let fields be added non-breakingly, but that itself is a breaking change for existing literals, so it's out of scope here — flagging only for awareness, not action.

Closes #307 syntax present. Great work.

PR #318, opened 2026-07-07
fix(flickering-cartpole): reject burst flicker_prob above burst_len/(burst_len+1)

### LINKED ISSUE #316: Burst-flicker: stationary blank rate deviates from p when flicker_prob > burst_len/(burst_len+1)
## Context

Found during review of PR #315 (issue #302). Minor edge case, not blocking.

`FlickeringCartPole::with_seed_probability_and_burst` docs state the stationary blank fraction equals `flicker_prob` **exactly**. The derivation is correct for the typical regime, but `advance_flicker` clamps the visible-to-blank switch probability to `[0, 1]`:

```rust
let mean_visible_run = mean_blank_run * (1.0 - self.flicker_prob) / self.flicker_prob;
let p_switch = (1.0 / mean_visible_run).clamp(0.0, 1.0);
```

When `flicker_prob > burst_len / (burst_len + 1)`, the derived mean visible-run length falls below 1 frame, `1/mean_visible_run > 1`, and the clamp engages. The chain then behaves as (visible always switches to blank), giving an actual stationary blank rate of `burst_len / (burst_len + 1)` instead of `p`.

Example: `p = 0.9`, `burst_len = 4` yields an actual blank rate of `4/5 = 0.8`, not `0.9`.

## Impact

- Default (`p = 0.5`, `burst_len = 4`) and all values tested (`p ≤ 0.7`) are unaffected.
- Behavior degrades gracefully (no panic, valid Markov chain), but silently contradicts the doc's "equals flicker_prob exactly" claim in the extreme-p regime.

## Suggested fix (either)

1. Tighten the constructor assert to also require `burst_len * (1 - flicker_prob) / flicker_prob >= 1.0` (i.e. `flicker_prob <= burst_len / (burst_len + 1)`), with a clear panic message, or
2. Document the achievable region and the clamped behavior in the constructor / module docs, plus a unit test at high `p`.

Option 1 is probably cleaner — a burst structure with mean visible runs shorter than one frame is not meaningful anyway.

rjwalters:
## Summary

In burst-structured flicker mode, `FlickeringCartPole::with_seed_probability_and_burst` derived a mean visible-run length of `burst_len * (1 - flicker_prob) / flicker_prob`. When `flicker_prob > burst_len / (burst_len + 1)` that length falls below one frame, `advance_flicker`'s `(1.0 / mean_visible_run).clamp(0.0, 1.0)` engages, and the visible state always switches to blank. The chain's stationary blank rate then silently settles at `burst_len/(burst_len+1)` instead of the requested `flicker_prob` (e.g. `p=0.9, burst_len=4` yields `0.8`), contradicting the docstring's "equals flicker_prob exactly" contract.

This implements the curator-recommended Option 1: a construction-time assertion that rejects the geometrically degenerate regime, so the documented contract holds across the entire accepted domain.

## Changes

- Add a third assertion in `with_seed_probability_and_burst` rejecting `flicker_prob > burst_len / (burst_len + 1)`, with a panic message reporting the computed limit (`{max_p:.4}`) and remediation guidance.
- Update the `# Panics` docstring to list all three panic conditions, including the new mean-visible-run constraint.
- Add `#[should_panic]` test `test_burst_prob_above_achievable_max_panics` (`p=0.9, burst_len=4.0`).
- Add edge-case test `test_burst_prob_at_achievable_max_does_not_panic` (`p=0.8, burst_len=4.0`, exactly at the limit).
- Extend `test_burst_mode_blank_rate_matches_p` to cover the boundary value `p=0.8`.
- No change to `advance_flicker` — the assert removes the reachable code path that triggered the clamp.

| Panics when `flicker_prob > burst_len/(burst_len+1)` | done | New assertion; `test_burst_prob_above_achievable_max_panics` passes |
| Panic message includes the computed limit | done | Message formats `burst_len/(burst_len+1) = {max_p:.4}` |
| `# Panics` lists all three conditions | done | Docstring updated to a three-item list |
| `#[should_panic]` test for the new condition | done | `test_burst_prob_above_achievable_max_panics` (expected = "burst mode: flicker_prob") |
| `cargo test --lib` passes, no regressions | done | 502 passed; 0 failed (training feature) |
| `cargo clippy --all-targets -- -D warnings` clean | done | `cargo clippy --all-targets --features training -- -D warnings` exit 0 |
| `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --all-features` clean | done | exit 0 |
| Existing `test_burst_mode_blank_rate_matches_p` still passes | done | Passes, now including boundary `p=0.8` |
| Edge case `p = limit` does not panic | done | `test_burst_prob_at_achievable_max_does_not_panic` passes |

## Test Plan

- `cargo test --lib --features training` — 502 passed, 0 failed, 3 ignored.
- `cargo clippy --all-targets --features training -- -D warnings` — clean.
- `cargo fmt -- --check` — clean.
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --all-features` — clean.

Closes #316

### DIFF
--- a/src/env/games/flickering_cartpole.rs
+++ b/src/env/games/flickering_cartpole.rs
@@ -213,15 +213,30 @@ impl FlickeringCartPole {
     ///
     /// # Panics
     ///
-    /// Panics if `flicker_prob` is not in the **open** interval `(0, 1)` (a
-    /// Markov burst structure is only meaningful with both states reachable) or
-    /// if `burst_len < 1.0` (a burst must last at least one frame).
+    /// Panics if any of the following hold:
+    /// - `flicker_prob` is not in the **open** interval `(0, 1)` (a Markov
+    ///   burst structure is only meaningful with both states reachable).
+    /// - `burst_len < 1.0` (a burst must last at least one frame).
+    /// - `flicker_prob > burst_len / (burst_len + 1)`. Beyond this limit the
+    ///   derived mean visible-run length `burst_len * (1 - flicker_prob) /
+    ///   flicker_prob` falls below one frame, which is geometrically
+    ///   ill-defined and would silently clamp the stationary blank rate to
+    ///   `burst_len / (burst_len + 1)` instead of `flicker_prob`. To reach a
+    ///   higher blank rate, increase `burst_len` rather than `flicker_prob`.
     pub fn with_seed_probability_and_burst(seed: u64, flicker_prob: f64, burst_len: f64) -> Self {
         assert!(
             flicker_prob > 0.0 && flicker_prob < 1.0,
             "burst mode requires flicker probability in (0, 1), got {flicker_prob}"
         );
         assert!(burst_len >= 1.0, "burst length must be >= 1.0, got {burst_len}");
+        let max_p = burst_len / (burst_len + 1.0);
+        assert!(
+            flicker_prob <= max_p,
+            "burst mode: flicker_prob {flicker_prob} exceeds the achievable maximum \
+             burst_len/(burst_len+1) = {max_p:.4} for burst_len {burst_len}; \
+             the mean visible-run length would be < 1 frame. \
+             Either reduce flicker_prob or increase burst_len."
+        );
         Self {
             inner: CartPole::new(),
             flicker_prob,
@@ -613,7 +628,9 @@ mod tests {
     fn test_burst_mode_blank_rate_matches_p() {
         // The whole point of the burst construction: the stationary blank rate
         // still equals p, so it is an apples-to-apples comparison with i.i.d.
-        for &p in &[0.3_f64, 0.5, 0.7] {
+        // `p = 0.8` sits exactly at the `burst_len/(burst_len+1)` limit for
+        // `burst_len = 4.0`, so it is the boundary case the constructor allows.
+        for &p in &[0.3_f64, 0.5, 0.7, 0.8] {
             let mut env = FlickeringCartPole::with_seed_probability_and_burst(123, p, 4.0);
             let stream = collect_flicker_stream(&mut env, 20000);
             let rate = stream.iter().filter(|&&b| b).count() as f64 / stream.len() as f64;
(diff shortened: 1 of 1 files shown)


rjwalters:
✅ **Approved!**

This is a clean, well-targeted fix for the burst-flicker stationary-rate deviation. The threshold math is exactly right and every acceptance criterion on #316 is met.

## Correctness — threshold math verified

The two-state Markov chain leaves *blank* with probability `a = 1/l_b` (`l_b = burst_len`) and leaves *visible* with probability `b = 1/l_v` where `l_v = l_b·(1-p)/p`. Stationary blank fraction is `b/(a+b) = l_b/(l_b + l_v) = l_b/(l_b/p) = p` — matches the docstring contract.

The visible→blank switch probability `1/l_v = p/(l_b(1-p))` exceeds 1 (triggering the `.clamp(0.0, 1.0)` in `advance_flicker`) exactly when `p > l_b/(l_b+1)`. The new assert rejects `flicker_prob > burst_len/(burst_len+1)`, so it eliminates the clamped regime precisely — no gap, no over-rejection.

**Boundary is handled correctly.** At `p = l_b/(l_b+1)`, `l_v = 1` exactly and `p_switch = 1.0`, so the clamp is a no-op and the stationary rate still equals `p`. The assert uses `<=`, so the boundary is (correctly) accepted rather than rejected.

## Floating-point boundary safety

The assert computes `max_p = burst_len/(burst_len+1.0)` while the clamp trigger derives from `mean_visible_run = burst_len·(1-p)/p < 1` — algebraically equivalent but different float expressions, so a sub-ULP discrepancy at the exact boundary is theoretically possible. It doesn't matter in practice: if `mean_visible_run` came out a hair below 1, `p_switch` would round to ~1.0 and clamp to 1.0, giving a rate deviation on the order of a ULP — unmeasurable. The assert removes the *meaningfully* clamped regime, which is the actual defect. For the tested case (`p=0.8`, `burst_len=4.0`), `4.0/5.0` and the literal `0.8` round identically in f64, so `0.8 <= 0.8` holds and the boundary test passes deterministically.

## Test coverage

- `test_burst_prob_above_achievable_max_panics` (`p=0.9, burst_len=4.0`) — `expected = "burst mode: flicker_prob"` matches only the new assert; the in-range `p`/`burst_len` guarantee it reaches the third assert, not the first two. Unambiguous.
- `test_burst_prob_at_achievable_max_does_not_panic` (`p=0.8`) — exercises the accepted boundary and checks `flicker_probability()`/`burst_length()`.
- `test_burst_mode_blank_rate_matches_p` extended with `p=0.8` — at the boundary the true rate is exactly 0.8, well inside the `< 0.05` tolerance over 20000 fixed-seed samples.

## Docs

The rewritten `# Panics` list cleanly enumerates all three conditions and the new entry explains *why* (mean visible-run < 1 frame) plus the correct remediation (increase `burst_len`, not `flicker_prob`).

## Verification

- Local: `cargo test --lib --features training flicker` → 22 passed, 0 failed (all three new/updated tests included).
- CI: all 9 checks pass (Tests macOS + Linux, Clippy, Documentation, Rustfmt, Security Audit, Publish dry-run, both Check-no-training).
- Merge state: CLEAN. PR description uses `Closes #316`.

Solid work.

